PR #71, opened 2026-08-04
fix(nemoclaw): pin mcp<2 so the MCP server starts

helenzhou0523:
## Problem

`integrations/nemoclaw/setup-tao-nemoclaw.sh` starts the host MCP server with an unpinned dependency:

```bash
setsid nohup uv run --with mcp --with uvicorn python "$SERVER" ...
```

`mcp` 2.0.0 removed `mcp.server.fastmcp`, which `server.py` imports. `uv` now resolves 2.x, so the server dies on import and bring-up fails:

```
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

`mcp` 1.19.0 still exposes the module.

## Why this was easy to miss

Hosts whose server process started before 2.0.0 was published keep running fine — the break only appears on a fresh install or a restart.

It also fails in a confusing way. The server launch is step 2 of `setup-tao-nemoclaw.sh` and skill installation is step 3, so a failed start leaves the sandbox holding whatever skills were installed previously, while the script reports an error about the server only.

## Change

Pin `mcp<2` at both call sites — the setup script and the run instructions in the `server.py` docstring — with a comment explaining the constraint.

## Verification

On a DGX Spark host (GB10, Ubuntu):

- Before: fresh `setup-tao-nemoclaw.sh` → `server failed to start`, with the `ModuleNotFoundError` above in `tao-mcp-server.log`.
- After: server binds the docker bridge and answers over it (`HTTP 406` from the sandbox, i.e. the server responded).
- `integrations/nemoclaw` test suite: **30 passed**.
- Full downstream check: an agent drove `tao_exec` staging plus `tao_run` train/evaluate/inference to completion against `nvcr.io/nvidia/tao/tao-toolkit:7.1.0-pyt`.


helenzhou0523:
@claude review

hqtran-nv:
@helenzhou0523 could you help resolving merge conflicts? We just synced from GitLab main to GH. Thank you!

christinayyw (approved):

PR #94, opened 2026-08-06
[TAO-2231][Feature] Add tao-analyze-gaps-od-map skill

arihantj-nv:

Wraps `gap_analysis object_detection` to score per-image, per-class AP50 against a ground truth and emit the weak-image set that drives mining. It is the stage that decides which images a DEFT object-detection loop tries to improve on.

**Features**
- Ships the reference ITS weak thresholds as defaults, so a run that says nothing gets the gates the reference pipeline used.
- Ships a spec template that holds every default, and `verify_object_detection_spec.py` to check a filled one before a container starts.
- Reports the gated classes and the `default_*` fallbacks, so the criteria behind a weak set are recoverable from the run's own output.

**Constraints**
- **`default_ap50_threshold` must be `0.0` unless you mean otherwise.** TAO DS defaults it to `0.5`, so a spec listing some classes in `weak_thresholds` and saying nothing else still gates every class it did not list — inflating the weak set and the mining budget derived from it. An image is weak if **any** class fails, so one ungated class can select images on its own.
- `input_format` is **lowercase** here (`kitti`/`coco`), unlike `analytics kpi_analyze` in the same container which requires uppercase. Both spellings are correct for their own stage.
- KITTI passes a directory for the annotation paths, COCO a json file.

**Design**
- The template is the only place a default lives, and the verifier runs against the artifact that will actually execute — the same shape as `tao-analyze-detection-kpi`.
- The spec is authored inside `results_dir`. The run writes four artifacts and does not retain the spec, so a finished run would otherwise tell you which images were selected but not why.

Weak-image selection is the input to mining, and mining is what a DEFT loop spends its GPU hours on. A wrong threshold does not fail — it selects a different set of images and everything downstream proceeds normally, so the error only shows up as a loop that does not improve.

### Related issues

Part of the DEFT object-detection skill set: #114, #96, #97 and #99. All of them and this MR add the same `images.tao_toolkit.data_services_od` key to `versions.yaml`; whichever merges second onward drops its duplicate.

**Goal:** confirm the weak-image set is exactly what the thresholds imply, on a fixture small enough that the answer is derivable on paper, and confirm the documentation stands alone for an agent that has never seen the skill.

**Setup:**

* _Fixture_: 4 images at 200x200, KITTI ground truth and KITTI inference labels, every box placed so each IoU is exactly 1.0 or exactly 0
* _Parameters_: `iou_threshold 0.5`, `conf_threshold 0.0`, gates `car 0.99 / bicycle 0.7`, all three `default_*` at 0.0
* _Agent + Prompt_: for the skill test, an agent was given this and nothing else — no commands, no container name, no spec format:

  ```
  I have object-detection ground-truth labels and a model's inference labels,
  and I want to find which images the model is weak on.

  ground truth labels:  <fixture>/gt
  inference labels:     <fixture>/pred
  images:               <fixture>/images
  classes I care about: car (needs AP50 at least 0.99), bicycle (needs AP50 at least 0.7)
  write results to:     <out>

  Find the appropriate skill, read it, and work out the invocation yourself.
  ```

## Test 1: gated classes only

```
Settings:
iou_threshold: 0.5
conf_threshold: 0.0
weak_thresholds:
  car:     {ap50: 0.99}
  bicycle: {ap50: 0.7}
default_ap50_threshold: 0.0
default_recall_threshold: 0.0
default_precision_threshold: 0.0
```

The answer should be

```
img2_car_miss.jpg
```

Why:
- `img1_car_hit` has car AP50 1.0, at or above its 0.99 gate, so it passes.
- `img2_car_miss` has car AP50 0.0, below 0.99, so it is weak.
- `img3_bike_hit` has bicycle AP50 1.0, at or above its 0.7 gate, so it passes.
- `img4_sign_only` passes on car. `road_sign` is not listed in `weak_thresholds`, so it falls back to `default_ap50_threshold`. The comparison is strict `<`, and no value is below 0.0, so an unlisted class cannot make an image weak — which is what setting the default to 0.0 is for.

## Test 2: the fallback threshold gates unlisted classes

```
Settings: as Test 1, but default_ap50_threshold: 0.5
```

The answer should be

```
img2_car_miss.jpg
img4_sign_only.jpg
```

Why:
- `img2_car_miss` is unchanged — still weak on car.
- `img4_sign_only` now also fails: `road_sign` is still unlisted, but the fallback is 0.5, and its AP50 of 0.0 is below that.
- `img1_car_hit` and `img3_bike_hit` contain no `road_sign` ground truth, so no `road_sign` metric exists for them and neither is affected.

This is the one setting that silently changes which images are selected. TAO DS defaults `default_ap50_threshold` to 0.5, so a spec that lists some classes in `weak_thresholds` and says nothing else still gates every class it did not list — inflating the weak set and the mining budget derived from it. An image is weak if it fails on **any** class, so one ungated class is enough to select it.

## Skill test (agent-driven)

Separately from the fixture tests above, an agent was given only a goal and the input paths —
no commands, no image name, no spec format — and asked to find and use this skill unaided.

It selected the right skill from the frontmatter description without ambiguity, built a correct
spec from `assets/default_object_detection.yaml`, and produced the expected result: **1 weak
image, `img2_car_miss`**, with `img4_sign_only` correctly *not* gated on its 100%-missed
`road_sign` class. That last case independently confirms the `default_ap50_threshold: 0.0`
behaviour is load-bearing rather than incidental.

It could not run the skill as documented. Following SKILL.md's instruction to resolve the
container from `versions.yaml`, it pulled ~30 GB and then failed on
`invalid choice: 'object_detection'`. It recovered only by brute-force enumerating subtasks
across every local image. The fixes in this MR come from that run:

- SKILL.md now states where the image is chosen that no published image carries the action, and
  gives the one-line probe to confirm before pulling.
- `skill_info.yaml` typed the annotation paths as `file`; KITTI takes a directory, which is what
  the run actually used.

Co-authored-by: Claude Opus 5

### DIFF
--- a/skills/data/tao-analyze-gaps-od-map/SKILL.md
+++ b/skills/data/tao-analyze-gaps-od-map/SKILL.md
@@ -75,8 +75,15 @@ warns when a fallback is above zero, since that gates classes you did not list.
 
 Run from the `tao-skills-external` repo root.
 
+**Write the spec into the results directory.** The run emits four artifacts and
+does not retain the spec, so a completed gap analysis otherwise cannot tell you
+which thresholds produced its weak set — and that weak set sizes the mining
+budget downstream. Keeping them together makes the selection criteria recoverable
+from the run alone.
+
 ```bash
-SPEC=/absolute/path/to/object_detection.yaml
+RESULTS_DIR=/absolute/path/for/this/run          # results_dir in the spec
+SPEC="$RESULTS_DIR/object_detection.yaml"        # spec lives beside its outputs
 RUN_ROOT=/absolute/path/that/contains/annotations/images/and/results
 GPU_COUNT=1
 
--- a/skills/data/tao-analyze-gaps-od-map/SKILL.md
+++ b/skills/data/tao-analyze-gaps-od-map/SKILL.md
@@ -67,7 +67,13 @@ SPEC=/absolute/path/to/object_detection.yaml
 RUN_ROOT=/absolute/path/that/contains/annotations/images/and/results
 GPU_COUNT=1
 
-DS_IMAGE="$(scripts/resolve_versions_key.py images.tao_toolkit.data_services)"
+# NO PUBLISHED IMAGE CARRIES `gap_analysis object_detection` YET. Verified against
+# nvidia/tao-toolkit:7.1.0-data-services, nvstaging ds:7.0.1-rc-200-x86 and
+# ds:2026.7.31-rc-14-multiarch — all three offer only {vcn_aoi, vlm_bcq, default_specs}.
+# Probe before pulling ~30 GB:
+#   docker run --rm "$DS_IMAGE" gap_analysis nosuch 2>&1 | tail -2
+# and confirm object_detection is in the printed choice list.
+DS_IMAGE=nvcr.io/nvstaging/tao/tao-toolkit-ds:7.0.1-rc-200-x86  # unpinned: repin once the action is published
 
 docker run --rm --gpus "$GPU_COUNT" --ipc=host --network=host \
   -v "$RUN_ROOT:$RUN_ROOT" \
@@ -89,7 +95,13 @@ docker info > /dev/null
 2. Resolve and pull the data-services image if needed:
 
 ```bash
-DS_IMAGE="$(scripts/resolve_versions_key.py images.tao_toolkit.data_services)"
+# NO PUBLISHED IMAGE CARRIES `gap_analysis object_detection` YET. Verified against
+# nvidia/tao-toolkit:7.1.0-data-services, nvstaging ds:7.0.1-rc-200-x86 and
+# ds:2026.7.31-rc-14-multiarch — all three offer only {vcn_aoi, vlm_bcq, default_specs}.
+# Probe before pulling ~30 GB:
+#   docker run --rm "$DS_IMAGE" gap_analysis nosuch 2>&1 | tail -2
+# and confirm object_detection is in the printed choice list.
+DS_IMAGE=nvcr.io/nvstaging/tao/tao-toolkit-ds:7.0.1-rc-200-x86  # unpinned: repin once the action is published
 docker image inspect "$DS_IMAGE" > /dev/null || docker pull "$DS_IMAGE"
 ```
 
--- a/skills/data/tao-analyze-gaps-od-map/SKILL.md
+++ b/skills/data/tao-analyze-gaps-od-map/SKILL.md
@@ -67,13 +67,7 @@ SPEC=/absolute/path/to/object_detection.yaml
 RUN_ROOT=/absolute/path/that/contains/annotations/images/and/results
 GPU_COUNT=1
 
-# NO PUBLISHED IMAGE CARRIES `gap_analysis object_detection` YET. Verified against
-# nvidia/tao-toolkit:7.1.0-data-services, nvstaging ds:7.0.1-rc-200-x86 and
-# ds:2026.7.31-rc-14-multiarch — all three offer only {vcn_aoi, vlm_bcq, default_specs}.
-# Probe before pulling ~30 GB:
-#   docker run --rm "$DS_IMAGE" gap_analysis nosuch 2>&1 | tail -2
-# and confirm object_detection is in the printed choice list.
-DS_IMAGE=nvcr.io/nvstaging/tao/tao-toolkit-ds:7.0.1-rc-200-x86  # unpinned: repin once the action is published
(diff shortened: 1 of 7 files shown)


skills/data/tao-analyze-gaps-od-map/references/skill_info.yaml
```diff
@@ -0,0 +1,22 @@
+network_arch: tao-analyze-gaps-od-map
+type: data
+container_image: nvcr.io/nvstaging/tao/tao-toolkit-ds:7.0.1-rc-200-x86  # unpinned: no published data-services image carries `gap_analysis object_detection` yet; repin once it ships
```
arihantj-nv: update the container. unpin this

skills/data/tao-analyze-gaps-od-map/SKILL.md
```diff
@@ -0,0 +1,119 @@
...
+---
+
+# TAO Analyze Gaps OD mAP
+
+Use this skill to run TAO Data Services object-detection gap analysis. The skill compares ground-truth and inference annotations, computes per-image per-class TP/FP/FN/AP50 metrics, and identifies weak images where any class metric falls below its threshold. It does not run inference; upstream steps must produce the inference annotations first.
+
+The container entrypoint is:
+
+```bash
+gap_analysis object_detection -e /absolute/path/to/object_detection.yaml
```
arihantj-nv, line 30: where do we update the yaml file

arihantj-nv: make sure this is in the experiment directory so that we can recover for logging.

skills/data/tao-analyze-gaps-od-map/references/skill_info.yaml
arihantj-nv: Done — the `# unpinned:` override is gone and the container is a managed pin:

```yaml
container_image: nvcr.io/nvstaging/tao/tao-dataservices:gapanalysis-02  # versions-key: images.tao_toolkit.data_services_od
```

The override existed because `gap_analysis object_detection` is absent from the release image. Rather than leave the pin outside the release-bump machinery, `versions.yaml` gains `images.tao_toolkit.data_services_od` for it — following the `video_clip` precedent of carrying a staging image with an interim note. `stamp_versions.py` now maintains it and CI's `--check` catches drift.

Verified before pinning: the image runs all five actions the OD pipeline uses, and reproduces this skill's fixture exactly (1 weak image, `img2_car_miss`, with `road_sign` correctly ungated).

skills/data/tao-analyze-gaps-od-map/SKILL.md
arihantj-nv: It had no answer before — the skill shipped a template and no way to check a filled one, so every mis-spelling reached the container. Now:

```bash
cp skills/data/tao-analyze-gaps-od-map/assets/default_object_detection.yaml "$SPEC"
# fill ground_truth_ann_path, inference_ann_path, images_dir, results_dir, kpi, input_format
python3 skills/data/tao-analyze-gaps-od-map/scripts/verify_object_detection_spec.py --spec "$SPEC"
```

`assets/default_object_detection.yaml` is the only place a default lives, so nothing can disagree with it, and the new `verify_object_detection_spec.py` checks the filled artifact — same shape as `tao-analyze-detection-kpi`.

It rejects what an agent-driven test of this skill actually got wrong: uppercase `input_format` (correct for `analytics kpi_analyze` in the same container, wrong here), relative or missing paths, and a `weak_thresholds` entry written as a bare number rather than a mapping. It also reports every gated class and the `default_*` fallbacks, and warns when a fallback is above zero — the setting that silently gates classes you did not list.

arihantj-nv: Good catch — it wasn't, and the run doesn't help you. `gap_analysis object_detection` writes four artifacts (`weak_images`, `box_gaps`, `image_metrics`, `gap_report`) and does **not** retain the spec, so a finished run told you which images were selected but not why. That matters beyond this stage: the weak set sizes the mining budget downstream.

Quick Start now authors the spec inside `results_dir`:

```bash
RESULTS_DIR=/absolute/path/for/this/run          # results_dir in the spec
SPEC="$RESULTS_DIR/object_detection.yaml"        # spec lives beside its outputs
```

and `verify` warns when it is anywhere else, naming the path it should be at. `verify` also prints the gated classes and fallbacks, so the selection criteria are recoverable from the run's own output.

Applied the same to the three sibling skills, since none of them retained its spec either — #97 (KPI: `conf_threshold`, `num_recall_points`, `ignore_sqwidth` all change what the numbers mean), #114 (mining: budget, allocation policy, rare-class list), and #96 (embeddings: parquets compared in a mining step must share `model`/`model_path`).

ddevendran-nvidia (approved):

PR #74, opened 2026-08-05
Migrate GitLab main → GitHub main

christinayyw:
Migrates the latest **GitLab main** (`tao-skills-external @ 2cebd259`, the GitLab release→main merge / MR!156) into GitHub main.

- **GitLab main is source of truth** for content + skill signatures.
- **GitHub-native preserved:** `.github/` (CI, workflows, hooks, templates), CONTRIBUTING.md, SECURITY.md.
- **Signatures:** 55 common `.oms.sig` are identical GitLab↔GitHub; the 2 public-only sigs (tao-run-on-local-docker, tao-run-platform) are kept.
- ⚠️ 2 docs still reference internal `gitlab-master.nvidia.com` links (from GitLab content) — flagging for review.
- DCO signed-off. Ready to merge once blossom CI passes.

PR #95, opened 2026-08-06
[TAO-2229][Feature] Add tao-mine-od-images skill

arihantj-nv:
Wraps `tmm unique_neighbor_matching` to mine images similar to a weak set from a large source pool, with class-stratified allocation and cross-iteration exclusion.

Documents the budget contract, the rare-class list, and the exclude-previous parquet that keeps successive iterations disjoint.

Migrated from GitLab. Opened against GitHub `main`; the two histories share no common ancestor after the content-import migration (#74), so this is the branch's net content as a single signed-off commit.

# Testing

The fixture uses small parquet files whose answer can be derived by hand. Used Claude to run the skill and compared the output to the expected answer.

Embeddings are 4-dimensional and placed on the unit axes, so every euclidean distance is exact.

**Source pool**
```
/fx/src/car_a.jpg     [1.00, 0.00, 0.00, 0.00]
/fx/src/car_b.jpg     [0.98, 0.02, 0.00, 0.00]
/fx/src/bike_a.jpg    [0.00, 1.00, 0.00, 0.00]
/fx/src/bike_b.jpg    [0.00, 0.98, 0.02, 0.00]
/fx/src/far.jpg       [-1.00, 0.00, 0.00, 0.00]
```

**Target pool** — standing in for the weak images gap analysis selected
```
/fx/tgt/weak_car.jpg   [1.00, 0.00, 0.00, 0.00]
/fx/tgt/weak_bike.jpg  [0.00, 1.00, 0.00, 0.00]
```

## Test 1: nearest neighbours across two targets

```
Settings:
desired_unique_count: 4
distance_metric: euclidean
allocation_policy: global
candidate_expansion_factor: 5
exclude_path: null
rare_class_list: ""
```

The answer should be

```
/fx/src/bike_a.jpg
/fx/src/bike_b.jpg
/fx/src/car_a.jpg
/fx/src/car_b.jpg
```

Why — euclidean distance from each target to every source:

```
                 weak_car   weak_bike
car_a              0.000       1.414
car_b              0.028       1.386
bike_a             1.414       0.000
bike_b             1.400       0.028
far                2.000       1.414
```

- A budget of 4 unique files across 2 targets resolves to the 2 nearest per target.
- `weak_car` takes `car_a` (0.000) and `car_b` (0.028).
- `weak_bike` takes `bike_a` (0.000) and `bike_b` (0.028).
- `far.jpg` is the farthest point from `weak_car` at distance 2.000, and ties for farthest from `weak_bike` at 1.414. It is never in either target's top 2, so it is not mined.
- The two selections are disjoint, so the union is exactly 4 unique files and the budget is met at 100% coverage.

The point of the fixture is that `car_b` and `bike_b` are *near* misses — close enough to be the second neighbour, far enough not to be exact — so the result distinguishes real nearest-neighbour retrieval from returning only exact matches.

## Skill test (agent-driven)

An agent was given only a goal and the input paths — no commands, no image name, no spec
format — and asked to find and use this skill unaided.

It reached the correct result: **`car_a`, `car_b`, `bike_a`, `bike_b`**, with the distance-2.0
outlier `far.jpg` excluded, at 100% coverage. It also correctly chose this skill over
`tao-mine-nearest-neighbors` by reasoning that only this one exposes `desired_unique_count`.

The defects it hit are fixed in this MR:

- SKILL.md resolved the container from `versions.yaml`, which yields the 7.1.0 release image.
  That image does not carry `tmm unique_neighbor_matching`, so the documented Quick Start is a
  ~30 GB pull that then fails. SKILL.md now names the working pin and shows how to probe.
- `skill-card.md` named the action `tmm unique_neighbors`, pointed at `mining_summary.txt`
  (which belongs to the sibling skill), described this skill as computing embeddings when it
  consumes them, and linked four reference files that do not exist.
- The per-iteration output was documented as `_iter_<N>`; the file written is
  `_iteration_<N>_topn_<K>`.


skills/data/tao-mine-od-images/references/skill_info.yaml
```diff
@@ -0,0 +1,38 @@
+network_arch: tao-mine-od-images
+type: data
+container_image: nvcr.io/nvstaging/tao/tao-toolkit-ds:2026.7.31-rc-14-multiarch  # unpinned: earliest published image carrying `tmm unique_neighbor_matching`; repin to versions.yaml once the action ships in a release image
```
arihantj-nv, line 3: unpin

arihantj-nv:
Superseded by #94, which now carries both this skill and `tao-analyze-gaps-od-map`.

Both need the new `images.tao_toolkit.data_services_od` key in `versions.yaml`, and that file can only gain it once — split across two MRs, whichever landed second would either conflict on the shared file or reference a key that does not resolve yet and fail `stamp_versions.py --check` until the other merged.

Nothing is lost: every commit, the testing notes and the agent-driven test report are carried over into #94, and mining now moves off its `# unpinned:` override onto the managed key.

arihantj-nv:
Superseded by #114 — same branch and same commits, re-opened as a **stacked** MR based on #94 rather than `main`.

GitHub refused to reopen this one (`Could not open the pull request`) because the branch was force-pushed while it was closed, so a new MR was the only route.

Stacking is what lets both skills keep their own ticket: both need `images.tao_toolkit.data_services_od`, `versions.yaml` can only gain that key once, and basing #114 on #94's branch means the key resolves and `stamp_versions.py --check` passes on both. GitHub retargets #114 to `main` automatically once #94 merges.

PR #96, opened 2026-08-06
[TAO-2502][Feature] Add tao-generate-image-embeddings skill

arihantj-nv:

Wraps `embedding image_embeddings` to turn a parquet of image filepaths into a parquet of embeddings — the corpus every nearest-neighbour mining step draws on.

**Features**
- CLIP or SigLIP, from a HuggingFace id, a local HF snapshot directory, or a TAO `.pth`/`.ckpt` checkpoint.
- Extra columns in the input parquet are carried through to the output verbatim, so metadata such as `label` survives into the embedding parquet.
- Ships a spec template holding every default and `verify_image_embeddings_spec.py` to check a filled one, including a write probe on the output directory so a run does not fail after doing the work.

**Constraints**
- **Every parquet compared against another in a mining step must come from the same `model` and `model_path`.** Embeddings from different encoders are not comparable, and mismatched encoders are the usual cause of mining output that looks unrelated to its targets. Nothing in the parquet records which encoder wrote it.
- `model_config_path` is required when `model_path` is a TAO checkpoint — a `.pth` or `.ckpt` cannot be rebuilt without the spec it was trained under.
- Needs HuggingFace access unless a local snapshot is supplied.
- Requires a `RUN_ROOT` bind mount with identical host and container paths, since the spec carries absolute host paths.

**Design**
- The template is the only place a default lives. The spec generator was removed: it loaded the template and then overwrote its values from argparse defaults unconditionally, so `--template` was inert for those fields. Here the defaults happened to agree with the template, so nothing was producing a wrong value — but it is the same two-sources-of-truth shape removed from the sibling skills.
- Validation runs on the artifact that will actually execute, regardless of how it was produced.
- The spec is authored beside `output_parquet`. This stage has no `results_dir` — `ImageEmbeddingsConfig` has six fields and emits one named parquet — so "beside its output" means that parquet's directory.

Mining compares embeddings, so the encoder is part of the experiment. Without the spec kept next to the parquet there is no record of which encoder produced it, and an encoder mismatch does not fail — it returns neighbours that are simply wrong.

### Related issues

Part of the DEFT object-detection skill set: #94, #114, #97 and #99. All of them and this MR add the same `images.tao_toolkit.data_services_od` key to `versions.yaml`; whichever merges second onward drops its duplicate.

**Goal:** exact embedding values cannot be derived by hand, so assert the geometry instead — byte-identical images must embed identically, and distance must increase with visual difference. A broken encoder or a mismatched preprocessing path breaks one or both.

**Setup:**

* _Fixture_: 4 solid-colour PNGs at 224x224. `red_a` and `red_b` are byte-identical (same md5), `red_c` is a near red, `blue` is far. The input parquet carries an extra `label` column to check pass-through.
* _Parameters_: `model SigLIP`, `model_path google/siglip-base-patch16-224`, `batch_size 64` — all from the template, unchanged

## Test 1: identical inputs embed identically, and distance tracks difference

The answer should be

```
rows 4, embedding dim 768, `label` column carried through

cosine distance   red_a <-> red_b   0.000000     byte-identical images
                  red_a <-> red_c   small        near red
                  red_a <-> blue    largest      different colour

and strictly:  d(red_a, red_b) < d(red_a, red_c) < d(red_a, blue)
```

Why:
- `red_a` and `red_b` are the same bytes, so any deterministic encoder must map them to the same vector — distance exactly zero, not merely small. A non-zero value here means non-determinism or a preprocessing path that depends on something other than pixels.
- `red_c` differs from `red_a` only in shade, `blue` differs in hue, so a working encoder must order them that way. The ordering is the property mining depends on; the absolute values are not meaningful on their own.
- The `label` column is not used by the stage, so its survival confirms the documented pass-through of extra input columns.

**Result:** `0.000000`, `0.011445`, `0.081339` — strictly increasing, 4 rows, 768 dimensions, `label` present.

Co-authored-by: Claude Opus 5


skills/data/tao-generate-image-embeddings/references/skill_info.yaml
```diff
@@ -0,0 +1,18 @@
+network_arch: tao-generate-image-embeddings
+type: data
+container_image: nvcr.io/nvstaging/tao/tao-toolkit-ds:7.0.1-rc-200-x86  # unpinned: held at the build this skill is validated against; repin once 7.1.0-data-services is confirmed to carry the action
```
arihantj-nv: update this to current one

skills/data/tao-generate-image-embeddings/scripts/verify_image_embeddings_spec.py
```diff
@@ -0,0 +1,37 @@
...
+
+"""Validate an existing TAO Data Services image-embeddings spec."""
+
+from __future__ import annotations
+
+import argparse
+import sys
+from pathlib import Path
+
+from prepare_image_embeddings_spec import load_yaml, validate_config
```
arihantj-nv: this is just a test so I think we can remove this. add to the prepare.

skills/data/tao-generate-image-embeddings/scripts/prepare_image_embeddings_spec.py
```diff
@@ -0,0 +1,142 @@
...
+"""Create and validate a TAO Data Services image-embeddings spec."""
+
+from __future__ import annotations
+
+import argparse
+import sys
+from pathlib import Path
+from typing import Any
+
+try:
```
arihantj-nv: we should assume we get a spec from the user, and then just verify that it works like in kpi analyze

ddevendran-nvidia (approved):

PR #97, opened 2026-08-06
[TAO-2501][Feature] Add tao-analyze-detection-kpi skill

arihantj-nv:

Wraps `analytics kpi_analyze` to score object-detection predictions against a KITTI or COCO ground truth, reporting per-class TP/FP/FN and AP50 plus PR curves. It is how a DEFT loop knows whether an iteration helped.

**Features**
- Ships a spec template holding every default, and `verify_kpi_analyze_spec.py` to check a filled one before a container starts.
- `verify` reports `conf_threshold`, `num_recall_points` and `ignore_sqwidth` — the three settings that change what the numbers mean — so a result is interpretable from the run's own output.
- Multiple `kpi_sources` in one run, each scored and reported separately.

**Constraints**
- `data.input_format` is **uppercase** here (`KITTI`/`COCO`), unlike `gap_analysis object_detection` in the same container which requires lowercase. Both are correct for their own stage.
- `data.mapping` values must be **lists**, not bare strings. `construct_category_map` iterates the value, so `- car: car` yields the class names `c`, `a`, `r`, every metric comes out zero, and the run still exits 0.
- `is_internal: true` drops every class except `person` and appends a Summary row, changing what the report means.
- `image_dir` must not end in `/`: the reported `Sequence Name` is `image_dir.split('/')[-2]`.
- Comparing two runs is only meaningful when both used the same `num_recall_points` and `ignore_sqwidth`; 11-point and 101-point interpolation give different AP for the same curve.

**Design**
- The template is the only place a default lives. The spec generator was removed: it loaded the template and then overwrote its values from argparse defaults unconditionally, so the template's values for those fields could never take effect — and three of them disagreed with it.
- Validation runs on the artifact that will actually execute, regardless of how it was produced, rather than at generation time on one of two paths.
- The spec is authored inside `results_dir`, since the run writes `kpi_calc.csv` and a PR curve but not the settings that produced them.

This is the metric the whole loop is judged on. A setting that silently changes what is being measured — a string where a list was expected, a different recall-point count between runs — produces numbers that look fine and are not comparable.

### Related issues

Part of the DEFT object-detection skill set: #94, #114, #96 and #99. All of them and this MR add the same `images.tao_toolkit.data_services_od` key to `versions.yaml`; whichever merges second onward drops its duplicate.

**Goal:** confirm the reported counts and AP are exactly what the precision/recall table implies, on a fixture whose interpolated AP is computable by hand, and confirm the documentation stands alone for an agent that has never seen the skill.

**Setup:**

* _Fixture_: 10 images, one `car` ground-truth box each, one prediction each at a distinct descending confidence. Five sit exactly on the box (IoU 1.0), five at IoU 0 — so the ranked precision/recall table, and the interpolated AP over it, are derivable on paper.
* _Parameters_: `iou_threshold 0.5`, `conf_threshold 0.3`, `num_recall_points 101` and `11`, `ignore_sqwidth 0`
* _Agent + Prompt_: for the skill test, an agent was given this and nothing else — no commands, no container name, no spec format:

  ```
  I have object-detection ground truth and a model's predictions. Score the model
  — I want per-class TP/FP/FN and AP50.

  images:              <fixture>/images
  ground truth labels: <fixture>/gt
  prediction labels:   <fixture>/pred
  the only class is `car`
  write results to:    <out>

  Find the appropriate skill, read it, and work out the invocation yourself.
  ```
* NOTE: this fixture found two defects in the container's `kpi.py` — average precision computed from transposed arguments, and undetected ground truth scored as a true positive at `conf_threshold 0`. Both are fixed in the image above; the expected values below are the corrected ones.

## Test 1: counts

```
Settings:
iou_threshold: 0.5
conf_threshold: 0.3
num_recall_points: 101
ignore_sqwidth: 0
```

The answer should be

```
TP=5  FP=5  FN=5  Pr=0.5  Re=0.5
```

Why:
- Five predictions overlap their ground-truth box exactly, so five are true positives.
- Five predictions have IoU 0 against the only box in their image, so five are false positives.
- Those five images still contain a ground-truth box that nothing matched, so five are false negatives.
- `Pr = 5/(5+5) = 0.5`, `Re = 5/(5+5) = 0.5`.

The recall denominator is the full ground-truth count (10), not the number matched.

## Test 2: average precision

The answer should be

```
num_recall_points=101  ->  AP = 0.4569
num_recall_points=11   ->  AP = 0.4831
```

Why — ranked by confidence, the running precision/recall is:

```
rank  cum TP  cum FP  precision  recall
  1      1       0      1.0000     0.1
  2      2       0      1.0000     0.2
  3      3       0      1.0000     0.3
  4      3       1      0.7500     0.3
  5      4       1      0.8000     0.4
  6      4       2      0.6667     0.4
  7      5       2      0.7143     0.5
  8      5       3      0.6250     0.5
  9      5       4      0.5556     0.5
 10      5       5      0.5000     0.5
```

Interpolated precision — the maximum precision at any recall at or above `r`:

```
r <= 0.3        ->  1.0000
0.3 < r <= 0.4  ->  0.8000
0.4 < r <= 0.5  ->  0.7143
r  > 0.5        ->  0        (recall never exceeds 0.5)
```

Averaged over the 101 sampled recall points: `(31x1.0000 + 10x0.8000 + 10x0.7143) / 101 = 0.4569`. Over 11 points the buckets fall differently and the same curve gives `0.4831`.

Raising `num_recall_points` from 11 to 101 is not cosmetic — it moves the reported AP, so a comparison across runs is only meaningful when both used the same value. The template pins it.

## Test 3: the confidence floor does not change what was matched

```
Settings: as Test 1, but conf_threshold swept over 0.3, 0.1, 0.01
```

The answer should be `TP=5 FP=5 FN=5` at every setting, since all ten predictions carry confidence 0.5 or above and none of them is filtered out.

Why this is worth asserting: matching and thresholding are separate steps. `conf_threshold` decides which predictions are considered, and it cannot change whether a ground-truth box was covered by one. A sweep that leaves every prediction above the floor must therefore leave the counts untouched — if they move, the threshold is affecting matching rather than filtering.

The template ships a threshold above zero. An undetected ground-truth box is carried through matching as `(t=1, p=0.0)`, so a floor of exactly `0.0` sits on the boundary of the comparison that separates a matched box from a missed one.

## Skill test (agent-driven)

An agent was given only a goal and the input paths — no commands, no image name, no spec
format — and asked to find and use this skill unaided.

It got a correct run on the first try with no guessing on the spec: it copied the shipped
template and filled only the `null`s, taking every tuning value verbatim from the template and
choosing only the four paths and `results_dir`. Output matched the fixture's hand-derived
answer exactly — **`TP=5 FP=5 FN=5 Pr=0.5 Re=0.5`**.

It reported the `conf_threshold` guidance as the skill's strongest section: unambiguous about
the value, and precise about *why* it must exceed zero. It reported the container as the
weakest — SKILL.md resolved it from `versions.yaml` while `skill_info.yaml` carried a
conflicting `# unpinned:` override, with no cross-reference between them. Probing settled it:
`analytics kpi_analyze` **is** present in the 7.1.0 release image, so the override was
unnecessary and this MR restores the managed pin.

Also fixed from that run: SKILL.md told the reader to run from the repo root and then called
`resolve_versions_key.py` with no path, but that script resolves `versions.yaml` from
`$TAO_SKILL_BANK_PATH` or `~/tao-skills-external` and never cwd, so the documented line fails
for any checkout elsewhere.

One issue it raised is **not** fixed here, because it is upstream: `kpi_calc.csv` carries no
class column, so per-class attribution requires capturing stdout. SKILL.md describes the file
as "one row per sequence per class", which is misleading in this build.

Co-authored-by: Claude Opus 5


skills/data/tao-analyze-detection-kpi/references/skill_info.yaml
```diff
@@ -0,0 +1,22 @@
+network_arch: tao-analyze-detection-kpi
+type: data
+container_image: nvcr.io/nvstaging/tao/tao-toolkit-ds:7.0.1-rc-200-x86  # versions-key: images.tao_toolkit.data_services
```
ddevendran-nvidia: You'll want to update this now that 7.0.1 has been released. There's a script to apply the current versions.yaml and update it, but I can't remember the name. You could also patch it manually

skills/data/tao-analyze-detection-kpi/scripts/prepare_kpi_analyze_spec.py
```diff
@@ -0,0 +1,191 @@
...
+
+import argparse
+import sys
+from pathlib import Path
+from typing import Any
+
+try:
+    import yaml
+except ImportError:
+    yaml = None
```
ddevendran-nvidia: don't you want to raise a RuntimeError here since yaml is required by this code?

skills/data/tao-analyze-detection-kpi/scripts/prepare_kpi_analyze_spec.py
```diff
@@ -0,0 +1,191 @@
...
+    config["kpi"]["is_internal"] = args.is_internal
+
+    validate_config(config)
+    return config
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--image-dir", required=True, action="append",
+                        help="KPI source image directory. Repeat once per source.")
```
ddevendran-nvidia: Does "Repeat once per source" mean that you can give a list of directories?

skills/data/tao-analyze-detection-kpi/scripts/prepare_kpi_analyze_spec.py
```diff
@@ -0,0 +1,191 @@
...
+    parser.add_argument("--inference-ann-path", required=True, action="append",
+                        help="Inference annotations. Repeat once per source, in the same order.")
+    parser.add_argument("--mapping", required=True, help="Class-mapping YAML.")
+    parser.add_argument("--results-dir", required=True, help="Output directory for kpi_calc.csv.")
+    parser.add_argument("--output-spec", required=True, help="Path where the generated YAML is written.")
+    parser.add_argument("--template", help="Optional YAML template. Defaults to assets/default_kpi_analyze.yaml.")
+    parser.add_argument("--input-format", default="KITTI", choices=sorted(VALID_FORMATS))
+    parser.add_argument("--platform", default="local", choices=sorted(VALID_PLATFORMS))
+    parser.add_argument("--tag", default=None)
+    parser.add_argument("--iou-threshold", type=float, default=0.5)
```
ddevendran-nvidia: Wouldn't it be better if this were in the default spec included with this skill?

skills/data/tao-analyze-detection-kpi/SKILL.md
```diff
@@ -0,0 +1,194 @@
...
+```bash
+python3 skills/data/tao-analyze-detection-kpi/scripts/prepare_kpi_analyze_spec.py \
+  --image-dir /absolute/path/kpi/images \
+  --ground-truth-ann-path /absolute/path/kpi/labels \
+  --inference-ann-path /absolute/path/results/inference/labels \
+  --mapping /absolute/path/mapping.yaml \
+  --results-dir /absolute/path/results/analyze_kpi \
+  --output-spec /absolute/path/specs/kpi_analyze.yaml \
+  --input-format KITTI \
+  --iou-threshold 0.5 \
```
ddevendran-nvidia: If the user specifies this, shouldn't it get passed here? Same with conf-threshold and the other parameters?

skills/data/tao-analyze-detection-kpi/scripts/prepare_kpi_analyze_spec.py
arihantj-nv: there are defaults in spec, skills nd in yaml, do we need all 3?

skills/data/tao-analyze-detection-kpi/skill-card.md
```diff
@@ -0,0 +1,53 @@
...
+### Deployment Geography for Use: <br>
+Global <br>
+
+## Known Risks and Mitigations: <br>
+Risk: Review before execution as proposals could introduce incorrect or misleading guidance into skills. <br>
+Mitigation: Review and scan skill before deployment. <br>
+
+Risk: The reported sequence name is derived from the image directory path rather than configured, so two KPI sources can silently collide under one name. <br>
+Mitigation: The bundled spec validator rejects a trailing slash on image_dir; lay out image directories so the second-to-last path component is the intended sequence identifier. <br>
+
```
arihantj-nv, line 24: allow a user to give a kpi analyze spec files?
or they can just give a paths, and a spec file will be generated with defaults. 

should it be reading default spec file and fill in the paths.

skills/data/tao-analyze-detection-kpi/SKILL.md
arihantj-nv: do not take in any command line params for generating specs.

ddevendran-nvidia (commented):
This is fine except for a possible bug with not passing through user-supplied parameters (such as kpi.conf)

skills/data/tao-analyze-detection-kpi/references/skill_info.yaml
```diff
@@ -0,0 +1,22 @@
+network_arch: tao-analyze-detection-kpi
+type: data
+container_image: nvcr.io/nvstaging/tao/tao-toolkit-ds:7.0.1-rc-200-x86  # unpinned: held at the build this skill is validated against; repin once 7.1.0-data-services is confirmed to carry the action
```
arihantj-nv: update this to 7.0.1 (stamp version)

skills/data/tao-analyze-detection-kpi/scripts/prepare_kpi_analyze_spec.py
arihantj-nv: removed this script

sfarhat-nvidia (approved):
lgtm

PR #98, opened 2026-08-06
[TAO-2500][Feature] Add tao-train-codetr skill

arihantj-nv:
Adds the Co-DETR skill covering training and inference.

The substantive piece is `inference.category_mapping`: Co-DETR folds source classes into target classes **at detection time** and then runs `apply_category_mapping_groupnms` — per-output-category soft-NMS after the merge. A post-hoc remap of the written labels cannot reproduce that, because by then the per-class NMS has already run against the unfolded classes.

Migrated from GitLab. Opened against GitHub `main`; the two histories share no common ancestor after the content-import migration (#74), so this is the branch's net content as a single signed-off commit.

# Testing

Used Claude to run the skill against a real COCO-trained Co-DETR checkpoint (ViT-Large, `dataset.num_classes: 80`) over 8 traffic images, and compared the output to what the documented `category_mapping` semantics predict. The images contain cars, trucks and traffic lights, so every behaviour below is visible in the class histogram of the emitted KITTI labels.

## Test 1: the availability probe is required

```
docker run --rm "$TAO_PYT_IMAGE" codetr --help
```

The answer should be that the console script is **absent** and the module fallback works:

```
codetr: cannot execute: Is a directory
python3 -m nvidia_tao_pytorch.cv.codetr.entrypoint.codetr inference -e <spec>   -> exit 0
```

Why: `nvcr.io/nvidia/tao/tao-toolkit:7.0.1-pyt` ships the Co-DETR package but does not register the console script, and a bare `codetr` resolves to a directory on `PATH`. A skill that assumes the console script fails on an image where the network is perfectly usable, so the two-step probe is load-bearing rather than defensive.

## Test 2: folding renames the mapped classes and drops the rest

```
Settings:
inference.conf_threshold: 0.3
inference.category_mapping:
  vehicle: ["car", "bus", "truck"]
```

The answer should be

```
without category_mapping:   traffic light 106,  car 38,  truck 21
with category_mapping:      vehicle 59
```

Why:
- `car` and `truck` are both listed under `vehicle`, so their 38 + 21 = 59 boxes are emitted as `vehicle`.
- `traffic light` appears in no group. Unmapped originals are dropped, so all 106 boxes disappear — folding is a filter as well as a rename, and a class omitted from the mapping is silently discarded.
- The emitted class names follow the *output* categories, so nothing downstream ever sees `car` or `truck`.

## Test 3: the post-fold merge depends on `soft_nms_enabled`

```
Settings: as Test 2, but conf_threshold: 0.05 to admit overlapping candidates
```

The answer should be

```
model.soft_nms_enabled: False  ->  vehicle 277     (= 146 car + 102 truck + 29 bus)
model.soft_nms_enabled: True   ->  vehicle 151
```

Why:
- Folding groups classes the detector confuses with each other, so one vehicle is often detected as both `car` and `truck`. Those are separate classes before the merge and per-class NMS leaves both standing.
- The instant they fold together they become two boxes of the *same* class, and only a merge run **after** the fold removes the duplicate.
- With `soft_nms_enabled: False` no such merge runs: the folded count is exactly the sum of its parts, 277, identical to what renaming the labels afterwards would produce.
- With `soft_nms_enabled: True` the post-fold soft-NMS removes 126 duplicates.

This is the test that distinguishes folding at inference from folding afterwards. The advantage the setting is chosen for only exists when the merge is switched on, and the schema default is off.

## Skill test (agent-driven)

An agent was given only a goal — run inference and get a single `vehicle` class with no
duplicate boxes — plus the image and checkpoint paths. No commands, no image name, no spec
format, and no instruction about how to obtain a class list.

It produced a correct run on the first attempt with zero iterations, and independently ran its
own control to check the skill's dedup claim: **43 boxes with `soft_nms_enabled: True` against
60 with it `False`**, reproducing the effect documented above at a different confidence
threshold without being told to look for it.

It reported the availability probe and the `soft_nms_enabled` warning as the highest-value
lines in the file — each would otherwise have cost a failed run or a silently wrong result.

The one input it had to reverse-engineer was the classmap: the skill required it, specified its
ordering and exact-case matching, and shipped neither a file nor a source. The agent recovered
the list by grepping the container for `toothbrush`. That is now fixed — `references/coco80_classmap.txt`
ships with the skill, extracted from the container's own `METAINFO['classes']`.

Remaining issues it raised, not yet addressed:

- The pointer to `nvidia_tao_pytorch/cv/codetr/experiment_specs/*.yaml` is dead — no `.yaml`
  exists under `cv/codetr/` in `7.0.1-pyt`.
- `contiguous_labels` is mentioned as a condition in one place and documented as absent from
  the schema in another.
- No single copy-pasteable `docker run` example; `$CODETR` is defined as a bare command that
  the skill itself establishes cannot be invoked directly.


skills/models/tao-train-codetr/SKILL.md
```diff
@@ -0,0 +1,347 @@
...
+| Field | Holds |
+|---|---|
+| `model.pretrained_backbone_path` | a backbone-only checkpoint initialising `model.backbone` |
+| `train.pretrained_model_path` | a full Co-DETR (or compatible DETR-family) checkpoint to fine-tune from |
+
+Backbones come from NGC:
+
+```bash
+ngc registry model download-version \
+    nvidia/tao/pretrained_dinov2_classification_imagenet:vit_large_patch14_dinov2 \
```
yuw-nv: is this the correct ptm?

.claude-plugin/marketplace.json
```diff
@@ -65,6 +65,7 @@
         "./skills/models/tao-finetune-nv-tesseract-ad-diffusion",
         "./skills/models/tao-finetune-nv-tesseract-forecasting",
         "./skills/models/tao-finetune-video-clip",
+        "./skills/models/tao-train-codetr",
```
yuw-nv: group tao-train-codetr with other model names

skills/models/tao-train-codetr/SKILL.md
arihantj-nv: updating it to hf model

arihantj-nv: Good catch — it was not correct, and I have removed it.

`vit_large_codetr` is **ViT-L/16**: the COCO checkpoint carries `backbone.patch_embed.proj.weight (1024, 3, 16, 16)`, which I read straight off the weights to confirm. `pretrained_dinov2_classification_imagenet:vit_large_patch14_dinov2` is ViT-L/**14**, so its `(1024, 3, 14, 14)` patch embedding cannot load into it.

The section now leads with `train.pretrained_model_path` and `zongzhuofan/co-detr-vit-large-coco` — the checkpoint this skill is actually verified against, which carries a trained backbone *and* trained heads. Backbone-only checkpoints keep a warning to check patch size against the target backbone first, rather than naming one I have not verified.

.claude-plugin/marketplace.json
arihantj-nv: Done — moved it from the end of the block (after the `tao-finetune-*` entries) to sit with the DETR family, immediately before `tao-train-deformable-detr` and near `tao-train-dino` / `tao-train-grounding-dino`.

yuw-nv (approved):
LGTM

PR #99, opened 2026-08-06
[TAO-2230][Feature] Add tao-run-deft-object-detection workflow skill

arihantj-nv:

This workflow skill runs the DEFT loop for object detection: it scores a zero-shot Grounding DINO baseline against a KPI set, finds the images the model is weak on, mines visually similar data from a large unlabelled pool, stages it as training data, fine-tunes, and re-scores — repeating until the iteration budget is spent.

**Workflow Features**
- Disk-backed state machine (`deft_state.json` + `loop_log.jsonl`), so a run resumes after an agent interruption or context compaction. `audit_deft_run.py` reads the run from disk and answers with the single safe next action rather than trusting prior assistant prose.
- Prepares the source pool once: Co-DETR pseudo-labels it, `annotations convert` turns that into COCO and ODVG, and `embedding image_embeddings` produces the corpus every iteration mines against.
- Delegates every stage to a leaf skill — `tao-analyze-gaps-od-map` (#94), `tao-mine-od-images` (#114), `tao-generate-image-embeddings` (#96), `tao-analyze-detection-kpi` (#97), `tao-train-codetr` (merged, #98), `tao-train-grounding-dino` (already on main). This skill owns only the glue between them.
- Generates every spec from TAO's own `default_specs` or Hydra schema and applies a version-controlled overlay, so no spec is hand-written and no stage silently inherits a TAO default the skill does not want.
- Mines with class-stratified allocation and cross-iteration exclusion, so successive iterations are disjoint.
- Autonomous after a single user gate at the Pre-Flight Summary.

**Workflow Constraints**
- **`captions` must list every class present in the KPI ground truth, not only the classes being trained.** Grounding DINO is open-vocabulary and assigns a detection to a class by the position of the caption token it matched, so a ground-truth class absent from `captions` can never be predicted — every one of its boxes is a permanent false negative, and its objects resurface as false positives on whichever caption the model matches instead. Nothing currently enforces this.
- **Co-DETR pseudo-labelling needs `model.soft_nms_enabled: True`.** The schema default is `False`, and with it off `inference.category_mapping` renames classes without merging, so an object detected as both `car` and `truck` yields two boxes of the same class in the training data. Measured at 277 boxes off versus 151 on.
- Grounding DINO training requires a validation source, and it must be **0-based** COCO — `dataloader/coco.py` uses `category_id` verbatim as a dense label index, so standard 1-based COCO overflows on the last class.
- Needs HuggingFace access for the SigLIP encoder unless a local snapshot is supplied.

**Workflow Design**

Follows the DEFT AOI workflow skill: state logged for auto-resume, loop mechanics separated into `references/*.md` with only the high-level shape in `SKILL.md`, completion criteria per stage, and preflight plus verification between components.

Other choices:
- **Specs are emitted, never authored.** `emit_default_spec.py` takes the spec from TAO — `<module> default_specs` for data-services actions, an `OmegaConf.structured()` schema dump for the PyTorch ones — and `apply_spec_overrides.py` applies `assets/overlays/<stage>.yaml`. A value that lives only in prose is a value nobody sets: `kpi.ignore_sqwidth` defaults to 0 in TAO against the reference pipeline's 40, and a run that never mentions it scores a different set of boxes with nothing reporting a difference. A `--set` colliding with an overlay key is an error rather than last-write-wins.
- **Container images resolve from `versions.yaml` at Pre-Flight**, the same way `tao-run-deft-aoi` does, via `scripts/resolve_versions_key.py`. That script now reads the `versions.yaml` of the bank it is part of rather than guessing `~/tao-skills-external`, so resolution works from a git clone and a plugin install alike without `TAO_SKILL_BANK_PATH` being set.
- **Glue scripts are named for the stage they feed** — `prepare_<what>_for_<stage>` — because three of them feed `train` and two feed `mining`, and the previous `build_`/`make_`/`compute_`/`merge_` names said nothing about the consumer.
- **Poll artifacts, not processes.** `await_stage.py` waits on files and `status.json`, never process names, and rejects dangling symlinks.

DEFT is one of TAO's main products and this is the object-detection arm of it. The loop was previously a set of notebooks and hand-driven commands; this makes it an agent-runnable skill with recorded state, so a multi-hour run survives interruption and its result is reproducible from the artifacts it leaves behind.

### Related issues

Depends on #94 (`tao-analyze-gaps-od-map`), #114 (`tao-mine-od-images`), #96 (`tao-generate-image-embeddings`) and #97 (`tao-analyze-detection-kpi`). #98 (`tao-train-codetr`) is merged. All four open MRs and this one add the same `images.tao_toolkit.data_services_od` key to `versions.yaml`; whichever merges second onward drops its duplicate.

Introduces a new workflow skill. Adds one key to `versions.yaml`.

#### Accuracy Reproducibility Test

**Goal:** reproduce the reference DEFT object-detection runs produced by the kratos pipeline on the same SanTomas ITS data — same miner (`tmm unique_neighbor_matching`), same KPI set, `multiplier=1`, 1 epoch per iteration — for both the 5,000- and 100,000-image pools. The reference is a per-iteration bundle of `mined_filepaths.csv`, `training_data_sources.csv` and `kpi_calc.csv`, so both the images each iteration selects and the resulting per-class AP50 are directly comparable.

**Results — 5,000-image pool.** Thirteen stages, no stage committed `status=error`, no manual intervention. The 100,000-image run is in progress and will be added here.

| | bicycle AP50 | car AP50 | person AP50 | mAP |
|---|---|---|---|---|
| baseline (zero-shot GDINO) | 0.6853 | 0.8618 | 0.7608 | **0.7693** |
| after iteration 1 | 0.6997 | 0.8724 | 0.7703 | **0.7808** |

One epoch of fine-tuning on the mined data raised mAP by **0.0115**, and every class improved.

Per-iteration mechanics:

| | weak images | mined | note |
|---|---|---|---|
| iter 1 | 10,891 / 14,393 (76%) | 4,941 | 98.8% of the pool |
| iter 2 | 10,362 | — | 59 pool images left, below the budget; loop stopped |

The pool the loop trained on: 78,392 boxes over 4,997 of 5,000 images — car 72,287, person 4,103, bicycle 2,002.

`FP` counts run into the millions in `kpi_calc.csv` because inference and KPI both run at `conf_threshold: 0.0` by the skill's overlay, which is deliberate — it lets the threshold be swept afterwards from a single inference pass. Precision at that threshold is not meaningful; AP is.

**Comparison against the kratos reference is not valid at 5k, in either half:**

* _Mining._ The reference selected all 5,000 pool images and this run selected 4,941, so the sets agree almost entirely by construction. The pool is smaller than one iteration's budget, so the miner never had a real choice to disagree about.
* _AP._ The reference label space has four classes including `road_sign`; this run has three. Grounding DINO is open-vocabulary and assigns a box by matched caption position, so removing a class redistributes its objects rather than dropping them — the two AP columns are not measuring the same task. The reference bundle was also produced on an unidentified data-services build.

Both are properties of the 5k fixture, not of the skill. The 100k run has a pool 20× the iteration budget and is the comparison that can actually fail.

The Co-DETR pseudo-labelling spec is held to the reference pipeline's own
(`codino_infer_spec_conf_0.yaml`): on identical inputs the overlay reproduces its output box-for-box, departing only in `conf_threshold` (0.3 rather than the reference file's threshold-sweep 0.0) and in folding to the run's target classes.

**Setup:**

* _Parameters_: 3 target classes (bicycle, car, person), max 2 iterations, 1 epoch, multiplier 1, learning rate 2e-5, AP50 gates `car 0.99 / bicycle 0.7 / person 0.7`, SigLIP embeddings
* _Checkpoints_: Grounding DINO `grounding_dino_swin_tiny_commercial_trainable_v1.1` from NGC as the zero-shot baseline and the fine-tuning start point; Co-DETR ViT-Large COCO-80 (`zongzhuofan/co-detr-vit-large-coco`) for pool pseudo-labelling
* _Dataset_: SanTomas ITS. 5,000-image source pool; KPI set of 14,393 images with KITTI ground truth. A 100,000-image pool was run separately.
* _Hardware_: 2x RTX A6000 (48 GB)
* _Images_: `nvcr.io/nvidia/tao/tao-toolkit:7.1.0-pyt` for Co-DETR and Grounding DINO; `nvcr.io/nvstaging/tao/tao-dataservices:gapanalysis-02` for gap analysis, mining, embeddings, conversion and KPI
* _Agent + Prompt_: Used `claude` and asked it to run the loop with nothing but the run parameters — no commands, no container names, no spec formats:

  ```
  Run the DEFT object-detection loop.
  Pool images:      /path/to/test_package_5000_bike1500_person1500_car2000/images
  Target classes:   bicycle, car, person
  workspace         /path/to/deft_od_test
  results dir       /path/to/deft_od_test/run_full
  KPI images        /path/to/kpi_v3_with_labels/images
  KPI ground truth  /path/to/kpi_v3_with_labels/labels
  KPI class mapping /path/to/kpi_class_mapping.yaml   # list of single-key dicts, values are lists
  epochs 1 · 2 GPUs · multiplier 1 · allocation class-balanced
  ```

  Everything else — the two container images, the zero-shot checkpoint, every spec, the mining budget, the weak-image thresholds — the skill resolved itself. The leaf skills were each tested the same way, given a goal and input paths only.
* NOTE: Let the skill fetch the Grounding DINO checkpoint and generate the pool embeddings from scratch rather than supplying either.

What has been verified:

**Mechanics.** The 749-assertion state machine suite passes — resume, the audit's next-action contract, the zero-weak-image early stop, and the guard that rejects a target class the pool cannot supply.

**Stage skills.** Each leaf skill was tested independently against fixtures small enough to derive the expected answer by hand, and separately by an agent given only a goal and input paths — no commands, no image name, no spec format — to check the documentation stands on its own. Details are in each MR.

**End-to-end.** The 5,000-image pool completed every stage without manual intervention and stopped correctly when the pool ran dry — numbers above. The 100,000-image run is in progress.

Co-authored-by: Claude Opus 5


skills/applications/tao-run-deft-object-detection/scripts/prepare_class_mappings_for_mining_data_prep.py
```diff
@@ -0,0 +1,206 @@
...
+merge. That matters: one object detected as both ``truck`` and ``car`` becomes two
+boxes of the same class the moment they fold together, and only a post-fold NMS
+removes the duplicate. Renaming labels afterwards cannot.
+
+**annotations convert does the KITTI->COCO fold** via ``kitti.mapping``::
+
+    labels2cat = {label: k for k, v in cat_map.items() for label in v}
+    mapped = labels2cat.get(row_p[0], None)
+    if not mapped:
+        continue
```
arihantj-nv, line 25: can we see if we can combined build_fold and and build_kpi

skills/applications/tao-run-deft-object-detection/scripts/fetch_gdino_checkpoint.py
```diff
@@ -0,0 +1,153 @@
...
+            "--zero-shot-checkpoint to skip the NGC path entirely (air-gapped runs)."
+        )
+
+    # `ngc download-version` errors out when --dest does not already exist; it will
+    # not create it. Failing here with a clear message beats an opaque CLI error.
+    dest.mkdir(parents=True, exist_ok=True)
+
+    cmd = ["ngc", "registry", "model", "download-version", f"{MODEL}:{version}",
+           "--dest", str(dest)]
+    print(f"$ {' '.join(cmd)}", file=sys.stderr)
```
arihantj-nv, line 91: this should come from the gdino train skill and we don't need this.

skills/applications/tao-run-deft-object-detection/scripts/align_token_usage.py
```diff
@@ -0,0 +1,347 @@
...
+`output_tokens`, `cache_read_input_tokens`, and `cache_creation_input_tokens`
+(plus the 5m/1h breakdown).
+
+This script runs after a loop (or any time you want updated numbers). For each
+stage entry in `loop_log.jsonl`, it sums the usage of every assistant message
+whose timestamp falls in `(prev_entry.ts, this_entry.ts]` (the first entry
+covers `[transcript_start, entry_1.ts]`), then writes a per-stage `tokens`
+field and updates `context_tokens` to the real context size at stage end.
+
+The original `loop_log.jsonl` is rewritten atomically (tmp + rename). Existing
```
arihantj-nv: let me check if this is used

skills/applications/tao-run-deft-object-detection/assets/gap_analysis_object_detection.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+# Per-class AP50 gates. An image is weak if ANY listed class falls below its gate.
+# Reference ITS defaults: `car` is abundant and already well learned so it is held
+# to a strict gate; the loop exists to lift the rare classes, which get looser ones.
+weak_thresholds:
+  car: {ap50: 0.99}
+  bicycle: {ap50: 0.7}
+  person: {ap50: 0.7}
+
+# THE SETTING THAT SILENTLY CHANGES WHICH IMAGES ARE SELECTED.
```
arihantj-nv: unnecessary detail comment

skills/applications/tao-run-deft-object-detection/scripts/apply_spec_overrides.py
```diff
@@ -0,0 +1,261 @@
...
+"""Set values in a spec emitted by ``emit_default_spec.py``.
+
+Completes the pair: ``emit_default_spec.py`` produces the canonical spec with
+``???`` on mandatory fields, this fills them, and no spec is ever hand-edited.
+
+    --overlay assets/overlays/kpi_analyze.yaml --set results_dir=/abs/out
+
+Two kinds of override, deliberately separated:
+
+``--overlay FILE``
```
arihantj-nv: make this a bit more obvious on what is happening. what is overlay doing?

arihantj-nv: can we make this --apply-workflow-defaults

skills/applications/tao-run-deft-object-detection/scripts/fetch_gdino_checkpoint.py
arihantj-nv: this is a bug in the gdino train skill. will file it. this is fine for now

skills/applications/tao-run-deft-object-detection/scripts/await_stage.py
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Block until a long stage finishes, by watching what it produces.
+
+**Never wait on a process name.** ``pgrep -f "grounding_dino train"`` matches the
+waiting shell's own command line — the pattern is *in* it — so the wait can never
+end. That cost 10h11m of wall clock on the first end-to-end run: training had
```
arihantj-nv: can simplify this comment - not as post mortem

skills/applications/tao-run-deft-object-detection/scripts/prepare_val_split_for_train.py
```diff
@@ -0,0 +1,133 @@
...
+
+"""Carve a validation COCO out of the prepared source pool.
+
+``grounding_dino train`` cannot run without a validation set — the data module
+subscripts ``val_data_sources["json_file"]`` unconditionally — so every iteration
+needs one. Deriving it from the pool means it always exists, always matches the
+target classes, and needs nothing from the user.
+
+**The category ids are rewritten to 0-based, and that is the whole point.**
+Grounding DINO's validation loader does::
```
arihantj-nv, line 13: we do not need this post mortem - just tell me what this class does

skills/applications/tao-run-deft-object-detection/scripts/prepare_class_mappings_for_mining_data_prep.py
arihantj-nv: cannot

ddevendran-nvidia (approved):

PR #101, opened 2026-08-06
Add GitHub Actions CI (static + platform nightly)

arif-ahmed-nv:
**Stacked on #100** — the nightly workflow runs the contract tests that PR adds, so this should merge second. Its diff will shrink to just `.github/workflows/` once #100 lands.

`.gitlab-ci.yml` does not run on GitHub, so after the migration every guard the old pipeline enforced silently stopped running. Two workflows port them.

## `skill-bank-static.yml` (PR + push to main)

Mirrors the GitLab `validate-skills` job: `validate-skills.sh`, `stamp_versions.py --check --strict-strays`, the NemoClaw integration tests, and `pytest scripts/tests templates/tests`.

`templates/tests` is included deliberately — it existed but was **absent from the GitLab job**, which is how the k8s multi-node doc/template drift shipped unnoticed.

## `platform-nightly.yml` (scheduled + manual dispatch)

Adds execution coverage, split by what each tier can actually prove. All three jobs run on GitHub-hosted CPU runners:

| Job | Proves |
|---|---|
| `contract` | Templates render; kubeconform accepts them offline |
| `smoke-docker-cpu` | Verb lifecycle + CLI syntax acceptance |
| `smoke-kubernetes` (kind) | A real API server **admits** the manifest |

`kubectl --dry-run=client` is not a substitute for kubeconform here: it contacts the API server to resolve API groups, so it validates nothing on a clusterless runner.

The kind job proves admission only, not scheduling — both k8s templates hardcode `nvidia.com/gpu`, so with no GPU capacity advertised a rendered pod never leaves `Pending`. Confirmed against minikube: `0/1 nodes are available: 1 Insufficient nvidia.com/gpu`.

## Why GPU and SLURM are deliberately not Actions jobs

This repo is public. Self-hosted runners on a `pull_request` trigger would let any fork PR execute arbitrary code on internal hardware, and a GitHub-hosted runner cannot reach an internal-only cluster login node regardless.

The repo already has the right mechanism: `blossom-ci.yml` is a **bridge, not an executor** — an authorized maintainer comments `/build`, and after a vulnerability scan `Job-trigger` runs `OPERATION: START-CI-JOB` against `secrets.CI_SERVER`, handing work to internal Jenkins and posting the log back. (`runs-on: blossom` resolves to runner `ci-server` — the bridge box, not a GPU host.)

So the GPU tier is a Jenkins job whose payload stays in-tree:

```
PLATFORM_SMOKE_GPU=1 bash scripts/ci/platform_smoke.sh docker
```

That keeps the test logic reviewable here while Jenkins stays a thin trigger. A nightly against `main` needs no blossom gating at all — it is already trusted code, so a Jenkins timer suffices.

## Notes

- Both workflow files carry SPDX headers; all commits are DCO signed off.
- `smoke-kubernetes` and the SLURM path in `platform_smoke.sh` are still stubs that report TODO rather than passing as covered. Docker is fully implemented and verified green.

### DIFF
--- a/.github/workflows/platform-nightly.yml
+++ b/.github/workflows/platform-nightly.yml
@@ -1,0 +1,137 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Nightly platform execution CI.
+#
+# The static workflow proves the platform skills SAY the right thing. Only
+# execution proves a real CLI ACCEPTS it: `brev exec <inst> -- <cmd> <args>` was
+# textually plausible and rejected by the actual CLI, which silently broke three
+# of Brev's four verbs. Two tiers:
+#
+#   contract — real tools, no cluster/credentials/GPU. Renders every vendored
+#              template and validates it (kubeconform offline; `kubectl
+#              --dry-run=client` is NOT a substitute, it contacts the API server
+#              to resolve API groups). Also asserts documented brev/kubectl
+#              invocations against the installed CLI's own signature.
+#
+#   smoke-cpu — live submit/status/logs/cancel on a CPU runner. This covers the
+#              verb LIFECYCLE (record open/mark, container create/inspect/logs/
+#              rm, manifest admission) and CLI syntax acceptance — the class the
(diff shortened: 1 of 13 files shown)


arif-ahmed-nv:
Superseded by #103 — same commits, rebranched to the repo's `dev/` naming convention.

PR #116, opened 2026-08-11
[TAO-0000][Feature] Allow sourcing credential env files; load them where they are consumed

arif-ahmed-nv:
> **Title needs a real ticket ID** — replace `TAO-XXXX` before merge or `pr-title.yml` will fail.

## Why

The bank forbade reading or sourcing any credentials file and documented shell export as the only supported way to supply credentials. That does not work on runtimes which don't inherit a parent shell's exports (notably Codex).

Credentials may now reach the session either by export **or** by sourcing a user-approved env file.

**Unchanged:** never ask for values in chat, never print or read back a credentials file, never pass a secret on argv, never write secret values into records, logs, or commits.

## Three constraints that fell out of the change

Each was verified by running it, not assumed:

1. **Env files are bare `KEY=value`.** Plain `source` on such a file sets a shell variable *without exporting it*, so `-e HF_TOKEN` forwards nothing (`env | grep -c '^HF_TOKEN=' -> 0`). The loader is always `set -a; source /path/to/.env; set +a`. The bare form also keeps the file usable with `docker --env-file`, which rejects `export`-prefixed lines.

2. **Agent shells don't persist environment between tool calls.** Sourcing in one call and checking in the next returns `UNSET`. So the loader runs inside the *same fenced block* as the command consuming the credential. **39 of 46** credential-consuming blocks now carry it; the 7 exceptions are a documented sibling-block run in `docker-runs.md` (first block carries it, a sentence says the rest assume it) and one Brev command whose credential expands on the remote instance.

3. **`source` executes arbitrary shell**, so unprompted loading is limited to `~/.tao/secrets.env` and `~/.config/tao/.env`. A repo- or workspace-relative `.env` is loaded only when the user points at it — a `.env` from an extracted workspace would otherwise be code execution plus `PATH`/`AWS_ENDPOINT_URL` poisoning.

## Pre-existing bugs fixed while auditing the credential paths

- **`tao-run-on-slurm` sidecar wrote an empty token.** `ssh $LOGIN "... \"\$HF_TOKEN\""` expanded on the remote login node, where the variable is unset. Confirmed with a stub SSH that it produced `export HF_TOKEN=`, so gated downloads inside SLURM jobs have been failing regardless of runtime. Now expands locally and is fed over stdin. **Worth a sanity check on a real cluster.**
- **Five `-e VAR="$VAR"` forms** exposed secrets in the process table; two instructed the user to type the token literally. All use the pass-through form now.
- **`.env.example`** shipped `export`-prefixed lines and claimed shell exports take precedence over the file — backwards for `set -a; source`, where the file wins. Its empty `NGC_KEY=` placeholder silently clobbered a correctly-exported key.
- **"never persists credential values"** overstated what the bank does: `docker login`, the enroot install, and the SLURM sidecar all write credentials after approval. Scoped to unprompted writes.

## Verification

- `scripts/validate-skills.sh` — passes
- `pytest scripts/tests` — 182 passed, 2 skipped
- Loader coverage re-scanned with a deterministic fence walker: 39/46, remainder justified above
- Zero `-e SECRET=` argv forms remain

## Reviewer notes

- `skills/platform/tao-run-on-docker/SKILL.md` is at 19,864 of the 20,000-char cap — 136 chars of slack. The duplicated host-identity preamble (lines ~149-157 and ~207-215) would buy ~430 chars back, but that's a content change unrelated to credentials and was left alone.
- The repo's secret-on-argv lint (`test_skill_command_hygiene.py`) matches `-e VAR=$` but not the quoted `-e VAR="$`, which is why these leaks passed CI. Widening it is a separate change.


ramanathan831 (approved):

christinayyw:

**Latest run:** [tao-sb-OSS-develop #166](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/166/) — `597e15ce` @ 2026-08-11 21:29 UTC

|---|---|---|---|---|
| `skills/applications/tao-run-deft-iaa` | — | `claude` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/166/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-claude/skill-eval-report.html) |
| `skills/applications/tao-run-deft-iaa` | — | `codex` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/166/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-codex/skill-eval-report.html) |

PR #117, opened 2026-08-11
[TAO-000][CI] validate-skills: changed-only per-skill checks (+ drop README count)

christinayyw:
Two changes to `scripts/validate-skills.sh`:

### 1. Per-skill checks scope to changed skills (the main fix)
Per-skill checks (**1c, 2, 3, 3b, 3c, 4, 5, 5b, 6, 7**) now validate only the skills this branch changed vs the base ref. Repo-wide checks (**1, 1a, 1b, 1d**) still run whole-tree. Falls back to whole-tree when no base ref is available (local run or `--all`).

**Why:** a pre-existing violation in an *unrelated* skill (e.g. `tao-run-automl/SKILL.md` at 20,115 > 20,000, merged via #84/#78) was failing **every** PR — including doc-only ones like this — because the check scanned the whole tree. Changed-only isolates a PR to what it touches, and a failure now maps to the author's own change.

Base ref: `$VALIDATE_BASE_REF`, else `GITHUB_BASE_REF`/`CHANGE_TARGET`, else `origin/main`.

### 2. Drop the hard-coded README skill-count check (from earlier)
README `bundles all 65 skills` → `bundles all skills`; removed 1d's README-count comparison (kept manifest version-parity).

### Tested
- changed-only, 0 skills changed → per-skill skipped → **pass** (unblocks this PR)
- changed-only, automl changed → **3b flags it** (introducer still caught)
- `--all` → whole-tree still catches the automl breakage

### Follow-ups (separate)
- Make `validate-skills` a **required** check so per-skill violations are caught at the introducing PR (root cause of how the automl file merged).
- Add a **nightly whole-tree** run (advisory) to surface drift on untouched skills.
- One-time cleanup: trim `tao-run-automl/SKILL.md` under the cap.
- CI: ensure blossom provides a base ref (fetch `origin/main` or pass `VALIDATE_BASE_REF`); otherwise it safely falls back to whole-tree.


scripts/validate-skills.sh
```diff
@@ -118,11 +117,11 @@ else
   ok "no asset-only skill directories"
 fi
 
-# ─── 1d. manifest version parity + README skill count ───────────────────────
```
arihantj-nv, line 121: what is the purpose of this test? I will say this test did help one thing because it caught that my skill was accidentally listed under daft instead of data.

christinayyw: @arihantj-nv most are static test to catch formatting issues and some checks are required by skill signing

hqtran-nv (approved):

arihantj-nv (approved):

PR #124, opened 2026-08-12
[TAO-2230][Bugfix] deft-od: container mounts, and fixing skill references

arihantj-nv:

Two code fixes to container mounts, and a clarify issue in the references.

**Mounts.** Containers see only what is mounted. `init_deft_state.py` recorded `config.extra_container_mounts`, but no documented `docker run` used it, so any stage whose inputs sat outside `$WORKSPACE` — the normal layout for KPI data, pool images and the encoder — failed at the first container. The recorded list was also incomplete, omitting the Co-DETR checkpoint and classmap, the pool images, and the encoder itself. And the encoder entry pointed at the HuggingFace *snapshot* directory, which holds only symlinks into a sibling `blobs/`, so the container saw dangling links and the loader reported a missing model file for a model that was present.

`preflight.md` now exports the list as `$EXTRA_MOUNTS` and every launch carries it; inside the workspace it expands to nothing and commands are unchanged.

**References.** The remaining changes correct places where a reference described something the code or the container does not do: a

As merged, following the documented commands verbatim fails at the first container whenever the data is not inside the workspace, and the KPI stage can discard its own result.

No behaviour change for a run whose inputs are all inside the workspace. Runs with external inputs go from failing to working.

Three 5,000-image runs and one 100,000-image run, each driven only by the skill's documentation.

* **Reproducible.** Prep from scratch produced 78,392 boxes — bicycle 2,002, car 72,287, person 4,103 — identical to an independent earlier run, and baseline mAP 0.7692955 against a previous 0.7693.
* **5k:** mAP 0.7693 → 0.7805, all three classes up; `--require-complete` exits 0 on the pool-exhaustion early stop with `pool_remaining=59`.
* **100k, both iterations:** mAP 0.7631 → 0.7871 → 0.7946; `--require-complete` exits 0 with all iterations complete. Stratified mining lifted bicycle's share of mined boxes from 0.76% to 1.60% and person's from 3.64% to 5.76% against a pool that is 95.6% car — and those two classes are the ones whose AP improved (+0.059, +0.045).
* State-machine suite 820/820; `validate-skills.sh` and `stamp_versions.py --check --strict-strays` clean; no documented `docker run` is left mounting only `$WORKSPACE`.

Co-authored-by: Claude Opus 5

### DIFF
--- a/skills/applications/tao-run-deft-object-detection/references/tao-analyze-detection-kpi.md
+++ b/skills/applications/tao-run-deft-object-detection/references/tao-analyze-detection-kpi.md
@@ -70,11 +70,6 @@ and `ignore_sqwidth: 40`, so delegating to it and applying this overlay agree. A
 overlay anyway: the agreement is a fact about the current versions, not a guarantee, and
 a spec that states its own scoring settings is auditable after the fact.
 
-`conf_threshold: 0.0` is safe on the pinned image. An undetected ground-truth box used to
-be carried as `(t=1, p=0.0)` against a `p >= conf_threshold` check, so zero scored every
-missed box as a true positive; unmatched ground truth now uses a `-1.0` sentinel and lands
-in FN at any threshold. On a build predating that fix, use a small positive value.
-
 ## Scoring at more than one confidence threshold
 
 Inference runs at `0.0`, so the labels carry every detection and this stage can score them
--- a/skills/applications/tao-run-deft-object-detection/references/tao-analyze-detection-kpi.md
+++ b/skills/applications/tao-run-deft-object-detection/references/tao-analyze-detection-kpi.md
@@ -105,10 +105,10 @@ Pass the narrowed file as `data.mapping`, not the user's original.
 
 ## Invocation
 
-**Launch it detached.** This stage runs 25-45 minutes on a 14k-image KPI set — the
-baseline is the slowest, since every detection survives `conf_threshold: 0.0` — and the mAP
-appears *only* on stdout. A foreground `docker run | tee` loses the number if the client
-dies, while the container keeps running — name the container, redirect to the log, and
+**Launch it detached.** Runtime scales with the KPI set, and the baseline is the
+slowest phase to score because every detection survives `conf_threshold: 0.0`. The mAP
+appears *only* on stdout, and a foreground `docker run | tee` loses it if the client
+dies while the container keeps running — name the container, redirect to the log, and
 wait on the log with `await_stage.py`:
 
 ```bash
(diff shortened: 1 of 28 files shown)


skills/applications/tao-run-deft-object-detection/references/tao-analyze-detection-kpi.md
```diff
@@ -63,33 +63,17 @@ kpi:
...
-So: delegate the invocation, but always apply
-`assets/overlays/kpi_analyze.yaml` over the resulting spec.
+### These values are the leaf skill's defaults too
+
+`tao-analyze-detection-kpi` ships the same `conf_threshold: 0.0`, `num_recall_points: 11`
+and `ignore_sqwidth: 40`, so delegating to it and applying this overlay agree. Apply the
+overlay anyway: the agreement is a fact about the current versions, not a guarantee, and
+a spec that states its own scoring settings is auditable after the fact.
+
+`conf_threshold: 0.0` is safe on the pinned image. An undetected ground-truth box used to
```
sfarhat-nvidia: What is the point of having this (saying what it used to do) in the skill? Does it help the agent?

skills/applications/tao-run-deft-object-detection/references/tao-analyze-detection-kpi.md
```diff
@@ -63,33 +63,17 @@ kpi:
...
+be carried as `(t=1, p=0.0)` against a `p >= conf_threshold` check, so zero scored every
+missed box as a true positive; unmatched ground truth now uses a `-1.0` sentinel and lands
+in FN at any threshold. On a build predating that fix, use a small positive value.
```
sfarhat-nvidia: In what situation would it be using a build predating the fix? Do you have a section that ensures the user/agent is using the correct/up-to-date version of the conatiner?

skills/applications/tao-run-deft-object-detection/references/tao-analyze-detection-kpi.md
arihantj-nv: Agreed — removed in 61bcee7a. The paragraph described the defect the value guards against rather than the value itself, and the "build predating the fix" case is unreachable: the data-services image is pinned in `versions.yaml` and resolved at Pre-Flight (`preflight.md` step 4), so the run cannot be on an older one. `conf_threshold: 0.0` is now justified only where it is set — it keeps the full precision-recall curve so a threshold can be swept afterwards from a single inference pass.

Same treatment applied to `grounding-dino.md`, which justified waiting on artifacts with the wall-clock cost of a past run.

skills/applications/tao-run-deft-object-detection/references/tao-analyze-detection-kpi.md
arihantj-nv: Agreed — removed in 61bcee7a. The paragraph described the defect the value guards against rather than the value itself, and the "build predating the fix" case is unreachable: the data-services image is pinned in `versions.yaml` and resolved at Pre-Flight (`preflight.md` step 4), so the run cannot be on an older one. `conf_threshold: 0.0` is now justified only where it is set — it keeps the full precision-recall curve so a threshold can be swept afterwards from a single inference pass.

Same treatment applied to `grounding-dino.md`, which justified waiting on artifacts with the wall-clock cost of a past run.

sfarhat-nvidia (approved):

PR #125, opened 2026-08-13
[TAO-0][Bugfix] NVBug 6596799: Validate the complete IAA DEFT config

ramanathan831:
## Reported issue

NVBug 6596799: the IAA DEFT config loader accepted unknown keys and silently coerced present invalid values. For example, `mining.topn: 0` became 5, `experiment.visualize: "not_a_number"` became true, and a numeric path became an invented filesystem path.

## Original reproduction

1. Load the shipped `deft_config.yaml`.
2. Mutate fields one at a time to zero, null, an empty string, a wrong type, or an unknown key.
3. Instantiate `IaaDeftConfig`.

Before this change, invalid values were accepted across many fields and the resulting in-memory configuration no longer matched the file.

## Root cause

The loader coerced and defaulted values before validating the YAML structure. Falsy-value fallback expressions also made a present invalid value indistinguishable from an absent optional key.

## Implementation

- Validate the complete nested config before any coercion.
- Reject unknown sections and keys.
- Enforce boolean, numeric, enum, list, path-string, and range contracts by full dotted key.
- Preserve defaults only for genuinely absent optional keys.

## Regression coverage

Added/updated `scripts/tests/test_iaa_deft_config.py` with representative invalid values, unknown keys, the shipped valid config, and the absent-key default case.

## Post-fix verification

The refreshed branch rejected all original examples by file and full key: zero/null/empty `mining.topn`, non-boolean visualization values, numeric paths, invalid pool mode, invalid query settings, and unknown keys. Removing optional `mining.topn` still produced default 5.

- `python -m pytest -q scripts/tests` — **274 passed, 2 skipped**
- focused config/container tests — **21 passed**

## Residual risk

This validates the skill-owned configuration boundary rather than launching a GPU loop, which is appropriate for this loader-only failure. NVBug 6596801 is a narrower overlapping enum fix and may need conflict resolution depending on merge order.

Validation now occurs once at the skill-owned YAML boundary, before construction can coerce or default any value. Every downstream IAA code path therefore receives the same validated structure, and absent optional values remain distinct from present invalid values.

The change does not enumerate only the reported bad literals, silently replace malformed values with defaults, or add test-only exceptions. It validates the full schema and reports the complete field path.

### DIFF
--- a/skills/applications/tao-run-deft-iaa/scripts/iaa_deft/config.py
+++ b/skills/applications/tao-run-deft-iaa/scripts/iaa_deft/config.py
@@ -7,8 +7,415 @@
 """Parsed configuration for an IAA CLIP DEFT experiment."""
 
 import json
+import math
 import os
 import yaml
+
+
+_MISSING = object()
+
+
+def _error(config_path: str, key: str, message: str) -> None:
+    raise ValueError(f"{config_path}: {key} {message}")
+
+
+def _mapping(root, key: str, source: str, config_path: str, *, required=True):
+    value = root.get(key, _MISSING)
+    qualified = f"{source}.{key}" if source else key
+    if value is _MISSING:
+        if required:
+            _error(config_path, qualified, "is required")
+        return {}
+    if not isinstance(value, dict):
+        _error(config_path, qualified, "must be an object")
+    return value
+
+
+def _known(mapping, allowed, source: str, config_path: str) -> None:
+    unknown = sorted(set(mapping) - set(allowed), key=repr)
+    if unknown:
+        _error(
+            config_path,
+            f"{source}.{unknown[0]}" if source else unknown[0],
+            "is not a recognized setting",
+        )
+
+
+def _value(mapping, key: str, source: str, config_path: str, *, required=False):
+    value = mapping.get(key, _MISSING)
+    qualified = f"{source}.{key}" if source else key
+    if value is _MISSING and required:
+        _error(config_path, qualified, "is required")
+    return value, qualified
+
+
+def _string(
+    mapping,
+    key: str,
+    source: str,
+    config_path: str,
+    *,
+    required=False,
+    allow_empty=False,
+    choices=None,
+) -> None:
+    value, qualified = _value(
+        mapping, key, source, config_path, required=required
+    )
+    if value is _MISSING:
+        return
+    if not isinstance(value, str) or (not allow_empty and not value.strip()):
+        suffix = "string" if allow_empty else "non-empty string"
+        _error(config_path, qualified, f"must be a {suffix}; got {value!r}")
+    if choices is not None and value not in choices:
+        _error(
+            config_path,
+            qualified,
+            f"must be one of {', '.join(sorted(choices))}; got {value!r}",
+        )
+
+
+def _boolean(mapping, key: str, source: str, config_path: str, *, required=False):
+    value, qualified = _value(
+        mapping, key, source, config_path, required=required
+    )
+    if value is _MISSING:
+        return
+    if not isinstance(value, bool):
+        _error(config_path, qualified, f"must be a boolean; got {value!r}")
+
+
+def _integer(
+    mapping,
+    key: str,
+    source: str,
+    config_path: str,
+    *,
+    required=False,
+    minimum=None,
+) -> None:
+    value, qualified = _value(
+        mapping, key, source, config_path, required=required
+    )
+    if value is _MISSING:
+        return
+    if not isinstance(value, int) or isinstance(value, bool):
+        _error(config_path, qualified, f"must be an integer; got {value!r}")
+    if minimum is not None and value < minimum:
+        _error(config_path, qualified, f"must be >= {minimum}; got {value!r}")
+
+
+def _number(
+    mapping,
+    key: str,
+    source: str,
+    config_path: str,
+    *,
+    minimum=None,
+    maximum=None,
+) -> None:
+    value, qualified = _value(mapping, key, source, config_path)
+    if value is _MISSING:
+        return
+    if (
+        not isinstance(value, (int, float))
+        or isinstance(value, bool)
+        or not math.isfinite(value)
+    ):
+        _error(config_path, qualified, f"must be a finite number; got {value!r}")
+    if minimum is not None and value < minimum:
+        _error(config_path, qualified, f"must be >= {minimum}; got {value!r}")
+    if maximum is not None and value > maximum:
+        _error(config_path, qualified, f"must be <= {maximum}; got {value!r}")
+
+
+def _validate_config(root, config_path: str) -> None:
+    if not isinstance(root, dict):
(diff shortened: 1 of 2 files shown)


christinayyw:

**Latest run:** [tao-sb-OSS-develop #177](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/177/) — `4c0781d8` @ 2026-08-13 19:43 UTC

|---|---|---|---|---|
| `skills/applications/tao-run-deft-iaa` | — | `claude` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/177/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-claude/skill-eval-report.html) |
| `skills/applications/tao-run-deft-iaa` | — | `codex` | ❌ Fail | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/177/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-codex/skill-eval-report.html) |


skills/applications/tao-run-deft-iaa/scripts/iaa_deft/config.py
```diff
@@ -7,10 +7,417 @@
...
+    _integer(training, "num_nodes", "training", config_path, minimum=1)
+
+    mining = _mapping(root, "mining", "", config_path)
+    _known(
+        mining,
+        {"topn", "knn_metric", "knn_batch_size", "history_aware", "recovery"},
+        "mining",
+        config_path,
+    )
+    _integer(mining, "topn", "mining", config_path, minimum=1)
```
vpraveen-nv, line 207: **`minimum=1` closes the fall-through but also removes the reporter's stated use case.**

NVBug 6596799's very first example is *"Set the number of neighbours to search to 0 to turn mining off. It becomes 5."* The reporter reads `topn: 0` as the way to disable mining. After this change `topn: 0` is rejected outright rather than silently becoming 5 — which is the right direction, but it leaves no documented way to express "disable mining".

Worth confirming: was `0` ever a supported disable switch, or did the reporter infer that? If it was never supported, a line in `deft_config.yaml` next to `topn` saying so would stop the next person filing the same bug. If it was, `minimum=0` plus an explicit branch in the mining stage is the fix.

skills/applications/tao-run-deft-iaa/scripts/iaa_deft/config.py
```diff
@@ -56,6 +463,7 @@ def __init__(
 
         with open(config_path) as f:
             _cfg = yaml.safe_load(f)
+        _validate_config(_cfg, config_path)
```
vpraveen-nv, line 466: **The `or <default>` expressions below are now dead, but leaving them keeps the defect one edit away.**

With `_validate_config` running first, `int(_mining.get("topn", 5) or 5)` can no longer coerce a falsy value — validation already rejected it. But the `or 5` idiom is still the house style throughout `__init__`, so the next field added *without* a matching entry in `_validate_config` silently regains exactly the behaviour this PR is fixing, and no test will notice.

Two options, either would hold the line: strip the now-redundant `or <default>` tails so an unvalidated field raises instead of defaulting, or add a test that walks every leaf key in the shipped `deft_config.yaml` and asserts the validator knows it. The second also directly encodes the bug's "29 of 50 settings" finding as a regression test.

skills/applications/tao-run-deft-iaa/scripts/iaa_deft/config.py
```diff
@@ -7,10 +7,417 @@
...
+        _error(config_path, qualified, "must be an object")
+    return value
+
+
+def _known(mapping, allowed, source: str, config_path: str) -> None:
+    unknown = sorted(set(mapping) - set(allowed), key=repr)
+    if unknown:
+        _error(
+            config_path,
+            f"{source}.{unknown[0]}" if source else unknown[0],
```
vpraveen-nv, line 39: **Only the first unknown key is reported.**

`_known()` sorts the unknown set and names `unknown[0]`, so a config with several typos surfaces them one run at a time. The bug measured 80 silently-accepted mutations across 29 settings; someone cleaning up a hand-edited config will round-trip once per typo.

Joining the whole sorted list into the message is a one-line change and makes the error proportional to the problem.

sfarhat-nvidia:
@ramanathan831 is this still relevant given the config dataclass I pushed?

skills/applications/tao-run-deft-iaa/scripts/iaa_deft/config.py
ramanathan831: Addressed by the typed dataclass contract in #198. `MiningSection.topn` declares `valid_min=1`, matching the PAS reference/notebook contract: zero is rejected instead of being silently converted to the default. Controls where zero genuinely means unlimited are modeled explicitly with `valid_min=0` and covered by `test_typed_runtime_retains_notebook_zero_means_unlimited_controls`. I reran the typed runtime/config reproduction on #198: 22 focused tests passed.

skills/applications/tao-run-deft-iaa/scripts/iaa_deft/config.py
ramanathan831: Addressed by #198 at the root abstraction layer. `PasDeftConfig` materializes typed dataclass fields and validates their metadata constraints; it no longer parses validated values with `or <default>`, so a future field cannot regain the falsy-coercion defect merely by omitting a parallel hand-written check. The focused typed runtime/config suite passes 22/22.

skills/applications/tao-run-deft-iaa/scripts/iaa_deft/config.py
ramanathan831: Addressed by #198. `_reject_unknown` computes and reports the complete sorted unknown-key set (`contains unknown key(s): ...`) for every supported nested section, with parameterized coverage across root, experiment, mining recovery, and caption-diversity paths. The focused typed runtime/config suite passes 22/22.

ramanathan831:
Revalidated against #198 after the dataclass update. #198 now fixes NVBug 6596799 at the shared typed-config layer: it rejects complete unknown-key sets, validates types/bounds/options from dataclass metadata before runtime consumption, eliminates the old falsy `or <default>` parsing path, and distinguishes controls where zero is an intentional unlimited value from `mining.topn`, whose PAS reference contract requires a positive value. Focused verification on #198: `python -m pytest -q scripts/tests/test_iaa_deft_config_materialization.py -k "typed_runtime or dataclass_schema or mining_spec_is_the_only"` → 22 passed. This older hand-written validator is therefore redundant and would create two schemas to maintain. Closing it in favor of #198.

ramanathan831:
Superseded by the production dataclass configuration contract in #198; all review threads have been answered and resolved.

PR #129, opened 2026-08-13
[TAO-0][Bugfix] NVBug 6596812: Record and materialize the approved GPU shape

ramanathan831:
## Reported issue

NVBug 6596812: the TAO spec hard-coded one GPU independently for train, evaluate, and inference, never compared the selection with visible hardware, and did not preserve host inventory with the run.

The report's grep showed three copies of `num_gpus: 1` / `gpu_ids: [0]` on an eight-GPU host. A partial edit could leave other actions unchanged, and results did not identify available versus selected GPUs.

## Root cause

GPU shape was duplicated in stage templates and detached from preflight inventory. Initial remediation also materialized only train/evaluate, leaving inference unresolved; revalidation caught and corrected that remaining path.

## Implementation

- Make all three template shapes mandatory placeholders.
- Require and record the visible GPU ID inventory.
- Reject selections outside that inventory.
- Materialize the one approved shape into train, evaluate, and inference.
- Preserve inventory and selection in approval/state/audit contracts.

## Regression coverage and exact verification

On the actual 8×GPU host, preflight recorded visible IDs `0..7`; selecting `0,2` produced `(num_gpus=2, gpu_ids=[0,2])` in all three actions. A GPU absent from inventory is rejected. Additional revalidation commit: `[Bugfix] GPU materialization skipped the inference action`.

- focused GPU/container tests — **7 passed** before amendment; amended full suite **260 passed, 2 skipped**

## Residual risk

The workflow does not automatically consume every visible GPU; it makes the selected subset explicit and reproducible. Capacity/performance suitability remains a preflight/user decision.

GPU selection is captured once from the approved visible-device set and then materialized into every GPU-bearing TAO action: train, evaluate, and inference. Host approval, rendered specs, and runtime device scope now derive from the same recorded shape.

The change does not hardcode a machine-specific count, infer only from one environment variable, or patch train alone. Explicit selected IDs remain supported and are validated against the host inventory.

### DIFF
--- a/skills/applications/tao-run-deft-iaa/scripts/prepare_deft_config.py
+++ b/skills/applications/tao-run-deft-iaa/scripts/prepare_deft_config.py
@@ -272,9 +272,10 @@ def materialize(args: argparse.Namespace) -> dict[str, Any]:
             "gpu_ids": container_gpu_ids,
         }
     )
-    tao["evaluate"].update(
-        {"num_gpus": args.num_gpus, "gpu_ids": container_gpu_ids}
-    )
+    for action in ("evaluate", "inference"):
+        tao[action].update(
+            {"num_gpus": args.num_gpus, "gpu_ids": container_gpu_ids}
+        )
 
     _atomic_yaml(config_dir / "deft_config.yaml", deft)
     _atomic_yaml(config_dir / "tao_spec.yaml", tao)
--- a/skills/applications/tao-run-deft-iaa/specs/tao_spec.yaml
+++ b/skills/applications/tao-run-deft-iaa/specs/tao_spec.yaml
@@ -15,11 +15,12 @@ train:
   siglip_loss_mask_mode: none  # none or attribute_match_ignore
   triplet_loss_weight: 0.5  # set > 0 to enable triplet loss
   triplet_margin: 0.2
-  # Set num_gpus/gpu_ids to the shape approved in the Pre-Flight Summary;
-  # per-GPU batch_size is fixed, so world size multiplies the effective batch.
-  num_gpus: 1
+  # prepare_deft_config.py materializes the one approved GPU shape into each
+  # action that the workflow uses; this template carries no competing default.
+  # Per-GPU batch_size is fixed, so world size multiplies the effective batch.
+  num_gpus: ???
   num_nodes: 1
-  gpu_ids: [0]
+  gpu_ids: ???
   optim:
     optimizer_type: adamw  # adamw, lamb, or sgd
     vision_lr: 1.0e-7
@@ -71,15 +72,15 @@ evaluate:
   checkpoint: google/siglip2-so400m-patch16-256
   batch_size: 32
   num_workers: 8
-  num_gpus: 1
-  gpu_ids: [0]
+  num_gpus: ???
+  gpu_ids: ???
 
 inference:
   checkpoint: siglip2-so400m-patch16-256
   batch_size: 32
   num_workers: 8
-  num_gpus: 1
-  gpu_ids: [0]
+  num_gpus: ???
+  gpu_ids: ???
 
 export:
   checkpoint: ???
--- a/scripts/tests/test_iaa_deft_gpu_config.py
+++ b/scripts/tests/test_iaa_deft_gpu_config.py
@@ -68,6 +68,8 @@ def test_selected_gpu_shape_is_materialized_once_and_inventory_is_recorded(tmp_p
     assert approval["visible_gpu_ids"] == list(range(8))
     assert spec["train"]["gpu_ids"] == [0, 2]
     assert spec["evaluate"]["gpu_ids"] == [0, 2]
+    assert spec["inference"]["gpu_ids"] == [0, 2]
+    assert spec["inference"]["num_gpus"] == 2
     assert template["train"]["num_gpus"] == "???"
     assert template["evaluate"]["gpu_ids"] == "???"
     assert template["inference"]["num_gpus"] == "???"
--- a/scripts/tests/test_iaa_deft_gpu_config.py
+++ b/scripts/tests/test_iaa_deft_gpu_config.py
@@ -1,0 +1,116 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Regression tests for IAA DEFT GPU inventory and selection."""
+
+import json
+import subprocess
+import sys
+from pathlib import Path
+
+import yaml
+
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+IAA_ROOT = REPO_ROOT / "skills/applications/tao-run-deft-iaa"
+PREPARE = IAA_ROOT / "scripts/prepare_deft_config.py"
+INITIALIZE = IAA_ROOT / "scripts/init_deft_state.py"
+PYT_IMAGE = "nvcr.io/nvidia/tao/tao-toolkit:7.1.0-pyt"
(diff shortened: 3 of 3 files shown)


christinayyw:

**Latest run:** [tao-sb-OSS-develop #181](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/181/) — `f4dfe5e4` @ 2026-08-13 20:09 UTC

|---|---|---|---|---|
| `skills/applications/tao-run-deft-iaa` | — | `claude` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/181/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-claude/skill-eval-report.html) |
| `skills/applications/tao-run-deft-iaa` | — | `codex` | ❌ Fail | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/181/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-codex/skill-eval-report.html) |


skills/applications/tao-run-deft-iaa/scripts/prepare_deft_config.py
```diff
@@ -325,6 +343,11 @@ def _parser() -> argparse.ArgumentParser:
     parser.add_argument("--training-epochs", default=1, type=int)
     parser.add_argument("--num-gpus", default=1, type=int)
     parser.add_argument("--gpu-ids", default="0")
+    parser.add_argument(
+        "--visible-gpu-ids",
```
vpraveen-nv: **`required=True` on a new flag is a breaking change to this script's CLI — worth an explicit sweep of callers.**

`references/preflight.md` is updated in this PR, which covers the agent-driven path. The other places that name `prepare_deft_config.py` — `eval.config`, `evals/evals.json`, `references/scripts-and-agents.md` — only describe it in prose, so nothing literally breaks there. But anyone with a saved command line, and any in-flight run being re-prepared, now hard-fails with an argparse error.

If that's intended (it forces the host inventory to be recorded, which is the point of the bug), it would be worth saying so in the PR body so it isn't mistaken for a regression during release triage.

skills/applications/tao-run-deft-iaa/scripts/init_deft_state.py
```diff
@@ -228,6 +228,17 @@ def _load_run_config(args: argparse.Namespace) -> dict:
...
+    if (
+        not isinstance(visible_gpu_ids, list)
+        or not visible_gpu_ids
+        or any(
+            not isinstance(item, int) or isinstance(item, bool) or item < 0
+            for item in visible_gpu_ids
+        )
+        or len(set(visible_gpu_ids)) != len(visible_gpu_ids)
+    ):
+        raise ValueError("approval.json visible_gpu_ids must be unique non-negative integers")
```
vpraveen-nv: **`init_deft_state.py` treats `visible_gpu_ids` as mandatory while `audit_deft_run.py` treats it as optional.**

The audit gates its new checks behind `if "visible_gpu_ids" in config or "visible_gpu_count" in config`, so a run recorded before this change still audits cleanly. Here the key is required unconditionally, so the same pre-existing `approval.json` raises.

That asymmetry matters right now because two other open PRs in this batch — #134 (NVBug 6598785) and #138 (NVBug 6603466) — are specifically about recovering runs that were interrupted mid-flight. If a run started before this merges can no longer be re-initialised, the recovery paths those PRs add won't reach it. Is a legacy `approval.json` expected to be re-preparable, or is restarting the run the intended answer?

skills/applications/tao-run-deft-iaa/specs/tao_spec.yaml
```diff
@@ -15,11 +15,12 @@ train:
   siglip_loss_mask_mode: none  # none or attribute_match_ignore
   triplet_loss_weight: 0.5  # set > 0 to enable triplet loss
   triplet_margin: 0.2
-  # Set num_gpus/gpu_ids to the shape approved in the Pre-Flight Summary;
-  # per-GPU batch_size is fixed, so world size multiplies the effective batch.
-  num_gpus: 1
+  # prepare_deft_config.py materializes the one approved GPU shape into each
+  # action that the workflow uses; this template carries no competing default.
+  # Per-GPU batch_size is fixed, so world size multiplies the effective batch.
+  num_gpus: ???
```
vpraveen-nv, line 21: **Confirm nothing consumes the shipped template directly.**

Making these `???` is the right way to kill the three competing defaults the bug names. The risk is anything that loads `specs/tao_spec.yaml` as shipped rather than the materialized copy under `config/` — it now gets the string `"???"` where it previously got an `int`.

`prepare_deft_config.py` overwrites all three actions before writing, so the run path is fine. Worth a quick check that no spec-drift or schema test in `scripts/tests/` type-checks the template in place.

vpraveen-nv (approved):
minor nitpicks.

skills/applications/tao-run-deft-iaa/scripts/prepare_deft_config.py
ramanathan831: Swept all repository references to `prepare_deft_config.py`. The only executable skill invocation is `references/preflight.md`, and it supplies `--visible-gpu-ids`; `eval.config`, `evals/evals.json`, and `references/scripts-and-agents.md` only name the script in prose. The regression helper also supplies it. I kept the flag required for new runs because defaulting it from `gpu_ids` would falsely describe the selected subset as the full host inventory. Compatibility for already-created approvals is handled at initialization instead (commit 6eba8ca).

skills/applications/tao-run-deft-iaa/scripts/init_deft_state.py
ramanathan831: Fixed in 6eba8ca. Initialization now mirrors the audit contract: approvals created after this feature validate and persist the full visible inventory, while pre-change approvals without that field remain resumable and do not fabricate `visible_gpu_ids`/`visible_gpu_count`. Added `test_pre_inventory_approval_remains_resumable_without_fabricating_inventory`; the targeted GPU/container suite passes 9/9.

skills/applications/tao-run-deft-iaa/specs/tao_spec.yaml
ramanathan831: Confirmed by a repository-wide caller sweep and strengthened at the runtime boundary in 6eba8ca. `prepare_deft_config.py` is the sole executable consumer of the shipped template and materializes train/evaluate/inference shapes. `init_deft_state.py` requires the run-local `results/config/tao_spec.yaml` and now validates all three action shapes agree, so an unresolved `???` cannot enter runtime. Added `test_unmaterialized_gpu_placeholder_cannot_reach_runtime`; targeted suite passes 9/9.

sfarhat-nvidia:
This seems a bit overkill. What is the problem with the default spec having 1 GPU? Then the user/skill can just update the field to use however many GPUs they would like for each step. If it requests more than what's available, won't the TAO container fail?

ramanathan831:
The default of 1 GPU remains valid and this PR does not force a run to consume every available GPU. The defect is that the approved shape was not validated or preserved: a request for N GPUs could cross preflight without N visible devices, and host IDs such as `2,3` were also being reused as TAO's container-local ordinals even though Docker renumbers the selected allocation to `0,1`. Letting the container fail discovers that only after pulls, dataset/config preparation, and run-state creation, and it cannot distinguish capacity from namespace errors. The branch now validates the requested count before launch, records host and dense container IDs separately, and retains compatibility for approvals created before inventory recording. So `num_gpus=1` still works unchanged; this makes any user-selected shape fail early and reproducibly rather than relying on a late container failure.

PR #130, opened 2026-08-13
[TAO-0][Bugfix] NVBug 6596814: Carry the trained model across IAA iterations by default

ramanathan831:
## Reported issue

NVBug 6596814: the advertised improvement loop accumulated data but defaulted `continual_model` to false, so every iteration restarted from the base model.

`grep -nE '^ *continual_(model|dataset):' specs/deft_config.yaml` showed `continual_model: false` beside `continual_dataset: true`. The documentation did not explain a deliberate restart, and QA measured materially different accuracy.

## Root cause

The shipped template and CLI retained a non-continual model default inconsistent with the loop semantics and documentation.

## Implementation

Set `continual_model` to true in the template and CLI default, and update preflight/train references to describe the default behavior. Explicit `--continual-model false` remains supported.

- Original grep now shows both continual flags true.
- A fresh parser process reports `continual_model=True`.
- Explicit false parsing still returns false.
- `python -m pytest -q scripts/tests/test_iaa_deft_defaults.py` — **1 passed**

## Residual risk

Existing initialized runs retain their immutable approved configuration; this changes only new-run defaults. Users can still explicitly choose restart-from-base behavior.

The authoritative shipped template now defaults continual training on, while the parser continues to preserve an explicit false. Each new round therefore consumes the preceding trained checkpoint unless the user intentionally opts out.

The change does not force the value at runtime or ignore an explicit user choice; it corrects the default at its source.

### DIFF
--- a/skills/applications/tao-run-deft-iaa/references/preflight.md
+++ b/skills/applications/tao-run-deft-iaa/references/preflight.md
@@ -196,7 +196,7 @@ hardware, pool size, and accumulated data can change this substantially.
 | GPU shape | `num_gpus=1`, `gpu_ids=0` |
 | mining | budget `10000`, top-N `25`, cosine distance |
 | history selection | enabled, replay fraction `0.20` |
-| continual behavior | dataset `true`, model `false` |
+| continual behavior | dataset `true`, model `true` |
 | visualization | contact sheets `true`, embedding plot `true` |
 | Hugging Face token forwarding | disabled; enable only when the approved model/environment requires it |
 | PyTorch image | `nvcr.io/nvidia/tao/tao-toolkit:7.1.0-pyt` | <!-- versions-key: images.tao_toolkit.pyt -->
--- a/skills/applications/tao-run-deft-iaa/specs/deft_config.yaml
+++ b/skills/applications/tao-run-deft-iaa/specs/deft_config.yaml
@@ -12,7 +12,7 @@ iteration:
 
 training:
   init_checkpoint: ""
-  continual_model: false
+  continual_model: true
   continual_dataset: true
   num_nodes: 1
 
--- a/scripts/tests/test_iaa_deft_defaults.py
+++ b/scripts/tests/test_iaa_deft_defaults.py
@@ -1,0 +1,34 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Regression tests for IAA DEFT workflow defaults."""
+
+import importlib.util
+import sys
+from pathlib import Path
+
+import yaml
+
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+IAA_ROOT = REPO_ROOT / "skills/applications/tao-run-deft-iaa"
+
+
+def _prepare_module():
+    path = IAA_ROOT / "scripts/prepare_deft_config.py"
+    sys.path.insert(0, str(path.parent))
+    spec = importlib.util.spec_from_file_location("prepare_deft_config", path)
+    module = importlib.util.module_from_spec(spec)
+    assert spec.loader is not None
+    spec.loader.exec_module(module)
+    return module
+
+
+def test_continual_model_is_the_template_and_cli_default():
+    config = yaml.safe_load((IAA_ROOT / "specs/deft_config.yaml").read_text())
+    assert config["training"]["continual_dataset"] is True
+    assert config["training"]["continual_model"] is True
+
+    parser = _prepare_module()._parser()
+    action = next(item for item in parser._actions if item.dest == "continual_model")
+    assert action.default is True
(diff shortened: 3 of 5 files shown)


christinayyw:

**Latest run:** [tao-sb-OSS-develop #184](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/184/) — `ded6a705` @ 2026-08-13 20:49 UTC

|---|---|---|---|---|
| `skills/applications/tao-run-deft-iaa` | — | `claude` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/184/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-claude/skill-eval-report.html) |
| `skills/applications/tao-run-deft-iaa` | — | `codex` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/184/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-codex/skill-eval-report.html) |


skills/applications/tao-run-deft-iaa/specs/deft_config.yaml
```diff
@@ -12,7 +12,7 @@ iteration:
 
 training:
   init_checkpoint: ""
-  continual_model: false
+  continual_model: true
```
vpraveen-nv, line 15: **What happens on iteration 1, now that this is the default path?**

`clip-train-eval.md` in this PR says the adapter "uses the prior iteration's normalized model state" in continual-model mode. Iteration 1 has no prior iteration. With the old default of `false` that branch was only reached when a user opted in; it is now the shipped path for every run, so first-iteration behaviour is on the critical path.

The new test asserts the template and the argparse default are both `True`, but nothing here exercises iter 1 with `continual_model=True`. If the adapter falls back to `init_checkpoint`/base weights when no prior state exists, a test pinning that would be worth having; if it raises, this default can't ship as-is.

skills/applications/tao-run-deft-iaa/references/preflight.md
```diff
@@ -196,7 +196,7 @@ hardware, pool size, and accumulated data can change this substantially.
 | GPU shape | `num_gpus=1`, `gpu_ids=0` |
 | mining | budget `10000`, top-N `25`, cosine distance |
 | history selection | enabled, replay fraction `0.20` |
-| continual behavior | dataset `true`, model `false` |
+| continual behavior | dataset `true`, model `true` |
```
vpraveen-nv, line 199: **This changes a measured number, so the run record should carry it.**

The bug quantifies the flip as 57.9% → 65.1% after 5 rounds. That makes `continual_model` a variable that changes the reported metric — the same argument NVBug 6596812 (PR #129) makes for recording the GPU shape, where 1 vs 4 vs 8 GPUs moved Rank-1 by a point.

PR #129 adds `visible_gpu_ids`/`visible_gpu_count` to the immutable approval + state config for exactly this reason. `continual_model` deserves the same treatment, otherwise results recorded either side of this merge aren't comparable and nothing in the run says which mode produced them.

sfarhat-nvidia:
Hey @ramanathan831 , this should be marked as not a bug. The same one was filed for the notebook and a discussion there was had that concluded that this set default is not an issue.

skills/applications/tao-run-deft-iaa/specs/deft_config.yaml
ramanathan831: Verified against the PAS reference notebook checkout (`tao-tutorials`, branch `seanf/pas-deft`). Its shipped `tutorials/deft/specs/deft_config.yaml` deliberately sets `continual_model: false`; the skill's existing iteration-1 branch (`if not cfg.continual_model or number == 1`) is consistent with that contract. Making continual-model mode the default here would change notebook semantics rather than fix a defect. Per the application owner’s top-level review, this PR is being closed as Not a Bug.

skills/applications/tao-run-deft-iaa/references/preflight.md
ramanathan831: Agreed that this is metric-affecting and would need immutable recording if it were changed. The authoritative PAS notebook, however, intentionally ships `continual_model: false`, and the application owner confirmed that default is not a bug. The correct resolution is to preserve the existing recorded `continual_model` field and close this behavioral-change PR, rather than add provenance for an unsupported default flip.

ramanathan831:
Revalidated against the authoritative PAS notebook checkout and the skill runtime. The notebook intentionally ships `training.continual_model: false` with `continual_dataset: true`, and iteration 1 intentionally has no prior trained checkpoint. Changing the default would alter the experiment policy and measured results rather than repair a root-cause defect. Per Sean’s application-owner review, marking this Not a Bug and closing the PR; both inline threads have been answered and resolved.

ramanathan831:
Closed as Not a Bug after verifying the intentional notebook default and first-iteration behavior.

PR #131, opened 2026-08-13
[TAO-0][Bugfix] NVBug 6597142: Align IAA preflight with its credential-file contract

ramanathan831:
## Reported issue

NVBug 6597142: IAA's skill contract, environment template, and public card promised never to open or source credential files, while its required preflight reference contained `source /path/to/.env` and told users to point a loader at one.

The report's grep/sed sequence showed the contradictory safety statements and executable source command in the same workflow. QA did not observe a leak, but the documentation gave mutually exclusive instructions.

## Root cause

The preflight reference retained a generic credential-loader example despite IAA's stricter environment-only contract.

## Implementation

Remove the env-file source command and loader wording. Missing `NGC_KEY` now instructs the user only to export it in the shell launching the agent.

- Exact contract/card/template promises remain present.
- The refreshed IAA preflight contains neither `source /path/to/.env` nor `point the loader`.
- Hygiene regression guards both forbidden phrases.
- `python -m pytest -q scripts/tests/test_skill_command_hygiene.py` — **2 passed**

## Residual risk

Other skills intentionally permit user-approved env files under the repository-wide policy. This change is scoped to IAA because only IAA publishes the stricter never-source promise.

All IAA preflight guidance now follows the skill contract that credentials must already exist in the process environment. Removing the contradictory loader instruction from every affected reference eliminates the unsupported execution path itself.

The change does not add another implicit dotenv loader, hide failures, or merely reword one example while leaving an executable sourcing snippet elsewhere.

### DIFF
--- a/scripts/tests/test_skill_command_hygiene.py
+++ b/scripts/tests/test_skill_command_hygiene.py
@@ -131,3 +131,13 @@ def test_no_banned_command_patterns():
         "marker on a line inside a shell fence silences a REAL defect; prose "
         "outside shell fences is already exempt and needs no marker:\n"
         + "\n".join(violations))
+
+
+def test_iaa_preflight_never_loads_credential_files():
+    preflight = (
+        REPO_ROOT
+        / "skills/applications/tao-run-deft-iaa/references/preflight.md"
+    ).read_text(encoding="utf-8")
+
+    assert "source /path/to/.env" not in preflight
+    assert "point the loader" not in preflight
--- a/scripts/tests/test_skill_command_hygiene.py
+++ b/scripts/tests/test_skill_command_hygiene.py
@@ -41,6 +41,7 @@ from pathlib import Path
 REPO_ROOT = Path(__file__).resolve().parents[2]
 SCAN_DIRS = ("skills", "templates", "scripts", "integrations")
 SCAN_SUFFIXES = {".md", ".yaml", ".yml", ".sh", ".py", ".config", ".json", ".txt"}
+IAA_SKILL_ROOT = REPO_ROOT / "skills/applications/tao-run-deft-iaa"
 
 SECRET_VARS = "NGC_KEY|NGC_API_KEY|HF_TOKEN|BREV_API_TOKEN|WANDB_API_KEY|ACCESS_KEY|SECRET_KEY|AWS_SECRET_ACCESS_KEY"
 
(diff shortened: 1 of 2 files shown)


christinayyw:

**Latest run:** [tao-sb-OSS-develop #182](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/182/) — `0d639135` @ 2026-08-13 20:10 UTC

|---|---|---|---|---|
| `skills/applications/tao-run-deft-iaa` | — | `claude` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/182/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-claude/skill-eval-report.html) |
| `skills/applications/tao-run-deft-iaa` | — | `codex` | ✅ Pass (⚠️ w/ self-correction) | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/182/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-codex/skill-eval-report.html) |


scripts/tests/test_skill_command_hygiene.py
```diff
@@ -131,3 +131,13 @@ def test_no_banned_command_patterns():
...
         + "\n".join(violations))
+
+
+def test_iaa_preflight_never_loads_credential_files():
+    preflight = (
+        REPO_ROOT
+        / "skills/applications/tao-run-deft-iaa/references/preflight.md"
+    ).read_text(encoding="utf-8")
+
+    assert "source /path/to/.env" not in preflight
```
vpraveen-nv: **Exact-string assertions won't hold the property the bug is about.**

These two checks pin the literal strings that exist today. `source ~/.env`, `source "$ENV_FILE"`, or the same line re-added to a different reference all pass.

This file already has `test_no_banned_command_patterns`, which is the right home: adding `set -a`/`source *.env` to that banned-pattern list would cover every skill in the bank rather than one file in one skill, and would catch the same contradiction if it appears in a new workflow. NVBug 6597142's own framing is that the rule has to hold across all four documents — a repo-wide pattern check is what makes that true by construction.

vpraveen-nv (approved):
minor nitpick.

scripts/tests/test_skill_command_hygiene.py
ramanathan831: Fixed in ca647a7. The exact-string assertions are replaced by an executable-command scan across the complete IAA/PAS skill. It rejects `source <anything>`, POSIX `. <anything>`, variable-based file paths, and `set -a`, so equivalent spellings and newly added files are covered. I intentionally scoped the invariant to IAA/PAS: the shared platform contract and several other skills explicitly permit user-approved env-file loading, so a bank-wide ban would be incorrect. `python -m pytest -q scripts/tests/test_skill_command_hygiene.py` passes 2/2.

PR #132, opened 2026-08-13
[TAO-0][Bugfix] NVBug 6597479: Bundle the IAA dataset rebuild implementation

ramanathan831:
## Reported issue

NVBug 6597479: executable dataset-build logic arrived inside mutable `meta.tar.gz` rather than with the versioned skill, so identical skill revisions could run different rebuild code.

`find . -name rebuild.py` returned no repository implementation, while IAA references executed `$DATASET_ROOT/rebuild.py` extracted from the archive, contradicting the self-contained workflow contract.

## Root cause

The workflow treated executable code in a data export as its dataset builder, outside repository review and hash binding.

## Implementation

- Bundle `scripts/iaa_deft/rebuild.py` with the skill.
- Execute the bundled implementation with explicit metadata/output roots.
- Ignore any legacy archive copy.
- Update data-layout, preflight, and stage ownership documentation.

## Regression coverage and exact verification

- Original search now finds one skill-owned rebuild implementation.
- The supplied `~/pas/meta.tar.gz` still contains a legacy `rebuild.py`, but references explicitly leave it unused.
- A synthetic export rebuilt images/captions without export code.
- The bundled verifier ran read-only against the real rebuilt dataset: **2,096,238 expected, 2,096,238 images, 2,096,238 captions; 2,000 sampled; 0 problems; VERIFY: PASS**.

## Residual risk

The full multi-gigabyte archive was not destructively rebuilt in place during revalidation; the exact bundled code was exercised synthetically and its verification path was run over the complete existing dataset tree.

The rebuild implementation is now versioned and hash-bound with the skill that invokes it; dataset archives contain data only. The user-facing workflow always executes the bundled implementation and verifies its output.

The change does not copy or trust an archive-supplied program at runtime, fall back to that program when the bundled copy fails, or weaken dataset verification.

### DIFF
--- a/scripts/tests/test_iaa_deft_rebuild.py
+++ b/scripts/tests/test_iaa_deft_rebuild.py
@@ -20,16 +20,19 @@ def test_bundled_rebuild_uses_export_metadata_without_export_code(tmp_path):
     metadata = tmp_path / "metadata"
     output = tmp_path / "dataset"
     metadata.mkdir()
+    decoy = metadata / "rebuild.py"
+    decoy.write_text("raise RuntimeError('export-owned code was executed')\n")
     for split in ("train", "val", "test"):
-        unique_name = f"{split}.jpg"
-        raw = output / "images_raw" / split / unique_name
+        unique_name = f"nested/{split}.jpg" if split == "train" else f"{split}.jpg"
+        image_path = f"source/{split}.jpg"
+        raw = output / "images_raw" / split / image_path
         raw.parent.mkdir(parents=True, exist_ok=True)
         raw.write_bytes(split.encode())
         rows = [
             {
                 "unique_name": unique_name,
                 "source_split": split,
-                "image_path": unique_name,
+                "image_path": image_path,
                 "caption": f"{split} caption",
             }
         ]
@@ -53,8 +56,72 @@ def test_bundled_rebuild_uses_export_metadata_without_export_code(tmp_path):
     )
 
     assert "VERIFY: PASS" in result.stdout
-    assert (output / "images/train.jpg").resolve() == (
-        output / "images_raw/train/train.jpg"
+    assert (output / "images/nested/train.jpg").resolve() == (
+        output / "images_raw/train/source/train.jpg"
     )
-    assert (output / "captions/train.txt").read_text() == "train caption\n"
-    assert not (metadata / "rebuild.py").exists()
+    assert (output / "captions/nested/train.txt").read_text() == "train caption\n"
+    assert decoy.read_text() == "raise RuntimeError('export-owned code was executed')\n"
+
+    subset_verify = subprocess.run(
+        [
+            sys.executable,
+            str(REBUILD),
+            "--metadata-root",
+            str(metadata),
+            "--out",
+            str(output),
+            "--workers",
+            "1",
+            "--splits",
+            "train",
+            "--verify-only",
+        ],
+        check=True,
+        capture_output=True,
+        text=True,
+    )
+    assert "VERIFY: PASS" in subset_verify.stdout
+
+
+def test_rebuild_refuses_metadata_paths_that_escape_owned_roots(tmp_path):
+    metadata = tmp_path / "metadata"
+    output = tmp_path / "dataset"
+    metadata.mkdir()
+    raw = output / "images_raw/train/source.jpg"
+    raw.parent.mkdir(parents=True)
+    raw.write_bytes(b"image")
+    (metadata / "train_pairs.json").write_text(
+        json.dumps(
+            [
+                {
+                    "unique_name": "../escape.jpg",
+                    "source_split": "train",
+                    "image_path": "source.jpg",
+                    "caption": "caption",
+                }
+            ]
+        )
+    )
+    (metadata / "train_list.txt").write_text("../escape.jpg\n")
+
+    result = subprocess.run(
+        [
+            sys.executable,
+            str(REBUILD),
+            "--metadata-root",
+            str(metadata),
+            "--out",
+            str(output),
+            "--workers",
+            "1",
+            "--splits",
+            "train",
+        ],
+        check=False,
+        capture_output=True,
+        text=True,
+    )
+
+    assert result.returncode != 0
+    assert "must be a non-empty relative path" in result.stderr
+    assert not (output / "escape.jpg").exists()
--- a/scripts/tests/test_iaa_deft_rebuild.py
+++ b/scripts/tests/test_iaa_deft_rebuild.py
@@ -1,0 +1,60 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Regression tests for the skill-owned IAA dataset rebuild."""
+
+import json
+import subprocess
+import sys
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+REBUILD = (
+    REPO_ROOT
+    / "skills/applications/tao-run-deft-iaa/scripts/iaa_deft/rebuild.py"
+)
+
+
+def test_bundled_rebuild_uses_export_metadata_without_export_code(tmp_path):
+    metadata = tmp_path / "metadata"
+    output = tmp_path / "dataset"
+    metadata.mkdir()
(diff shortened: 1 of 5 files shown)


christinayyw:

**Latest run:** [tao-sb-OSS-develop #185](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/185/) — `96c140ff` @ 2026-08-13 21:00 UTC

|---|---|---|---|---|
| `skills/applications/tao-run-deft-iaa` | — | `claude` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/185/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-claude/skill-eval-report.html) |
| `skills/applications/tao-run-deft-iaa` | — | `codex` | ❌ Fail | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/185/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-codex/skill-eval-report.html) |


scripts/tests/test_iaa_deft_rebuild.py
```diff
@@ -0,0 +1,60 @@
...
+        capture_output=True,
+        text=True,
+    )
+
+    assert "VERIFY: PASS" in result.stdout
+    assert (output / "images/train.jpg").resolve() == (
+        output / "images_raw/train/train.jpg"
+    )
+    assert (output / "captions/train.txt").read_text() == "train caption\n"
+    assert not (metadata / "rebuild.py").exists()
```
vpraveen-nv: **This assertion is vacuous — the test never creates `rebuild.py`.**

`metadata` is a fresh `tmp_path / "metadata"` populated with only `*_pairs.json` and `*_list.txt`, so this can never fail and proves nothing about the behaviour in the test's own name ("without export code").

To actually pin NVBug 6597479, write a decoy `rebuild.py` into `metadata/` that would raise or write a sentinel if executed, then assert the sentinel is absent after the run — that demonstrates the bundled copy ran and the archive copy did not, which is the property the bug asks for.

skills/applications/tao-run-deft-iaa/scripts/iaa_deft/rebuild.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def _shard(rows, out_root: Path):
+    """Create one relative image symlink and caption for each pair row."""
+    made_links = made_caps = skipped = 0
+    for unique_name, source_split, image_path, caption in rows:
+        link = out_root / "images" / unique_name
+        target = Path("..") / "images_raw" / source_split / image_path
+        try:
+            link.symlink_to(target)
```
vpraveen-nv, line 36: **Breaks if `unique_name` ever contains a path separator.**

`link` is `out_root/images/<unique_name>` and `target` is the relative `../images_raw/<source_split>/<image_path>`. Both assume `unique_name` is a bare filename. If it carries a subdirectory, `link.parent` doesn't exist so `symlink_to` raises `FileNotFoundError` (only `FileExistsError` is caught), and the `..` in the target would resolve one level short.

The test only uses flat `train.jpg`/`val.jpg`/`test.jpg` names. If flat names are guaranteed by the exporter, a `parser.error` on the first name containing `os.sep` would make the assumption explicit and fail early instead of part-way through a 679k-row rebuild.

skills/applications/tao-run-deft-iaa/scripts/iaa_deft/rebuild.py
```diff
@@ -0,0 +1,138 @@
...
+def verify(splits, metadata_root: Path, out_root: Path, sample: int = 2000) -> bool:
+    """Check exact counts, then sample-resolve rebuilt links and captions."""
+    expected = 0
+    names = []
+    for split in splits:
+        names_for_split = (metadata_root / f"{split}_list.txt").read_text().split()
+        expected += len(names_for_split)
+        names.extend(names_for_split)
+
+    image_count = len(list((out_root / "images").iterdir()))
```
vpraveen-nv: **`--splits` subsets will fail verification against a shared flat output directory.**

`expected` is summed only over the requested splits, but `images/` and `captions/` are flat and shared across all splits. Running `--splits train` against a tree that already holds val/test entries counts every file and reports `VERIFY: FAIL` on a correct rebuild.

Since `VERIFY: PASS` is documented as the hard gate for the stage, that's a false stop. Counting only the names belonging to the requested splits would make the check split-aware.

skills/applications/tao-run-deft-iaa/references/data-layout.md
```diff
@@ -17,12 +17,14 @@ The IAA TAO-FT export contains:
...
 | `images_raw.tar` | yes | source images under `images_raw/<source_split>/...` |
-| `meta.tar.gz` | yes | pair/list metadata, README/vocabulary files, and `rebuild.py` |
+| `meta.tar.gz` | yes | pair/list metadata and README/vocabulary files |
 | `SHA256SUMS` | no | checksums for the two archives |
 
 There is no download branch. Preserve the archives in place; extraction goes
 to the approved `DATASET_ROOT`. Do not copy multi-gigabyte archives into the
-dataset tree.
+dataset tree. If an export also contains a legacy `rebuild.py`, leave it
+unused; the workflow always executes the hash-bound copy bundled with this
```
vpraveen-nv: **"hash-bound" isn't backed by anything in this PR.**

Moving `rebuild.py` into the repo does solve the substance of NVBug 6597479 — it's now git-versioned and reviewable, which the tarball copy never was. But the bug specifically points at the existing machinery: *"images and Python packages are pinned in versions.yaml with a CI check that fails if anything drifts."*

There's no `versions.yaml` entry or hash check added here, so "hash-bound" overstates it. Either wire it into the existing drift check, or soften the wording to "the copy bundled with this skill" so the doc doesn't promise a guarantee the repo doesn't enforce.

scripts/tests/test_iaa_deft_rebuild.py
ramanathan831: Fixed in b76a9df. The reproduction now places a hostile export-owned `metadata/rebuild.py` that raises immediately if executed, runs the bundled implementation, verifies PASS, and confirms the decoy remained untouched. This directly pins the ownership boundary instead of asserting absence of a file the test never created.

skills/applications/tao-run-deft-iaa/scripts/iaa_deft/rebuild.py
ramanathan831: Fixed in b76a9df. Nested `unique_name` values now preserve their relative hierarchy under both `images/` and `captions/`; link parents are created and the source target is computed relative to the actual link parent. All metadata paths are also validated as non-empty relative paths so `..`/absolute escapes fail before writing. Coverage includes `nested/train.jpg` and a traversal rejection.

skills/applications/tao-run-deft-iaa/scripts/iaa_deft/rebuild.py
ramanathan831: Fixed in b76a9df. Verification now compares requested image/caption path sets as subsets for partial `--splits` runs, while retaining exact-set verification for a complete train/val/test rebuild. The regression performs a full rebuild and then successfully reruns `--splits train --verify-only` against the shared output. Rebuild tests pass 2/2.

skills/applications/tao-run-deft-iaa/references/data-layout.md
ramanathan831: Corrected in b76a9df. This PR moves the executable into version-controlled, reviewed skill content but does not itself bind that file into a runtime digest, so the documentation no longer says “hash-bound.” I intentionally avoided inventing a redundant versions.yaml entry or a misleading hash claim; separate bundle-content binding remains the responsibility of the run approval/bundle contract.

ramanathan831:
Closing as Not a Bug per the PAS application-owner clarification. The bundled rebuild implementation in this branch generalizes behavior that was specific to the dataset delivered to Verkada, so it is not part of the supported PAS/IAA skill contract. The durable user-facing direction is to describe and validate the expected dataset layout transparently, matching the notebook flow, rather than ship the customer-specific rebuild path. Any documentation/layout-alignment change should be made in the canonical PAS contract branch, not retained as this bug fix.

PR #133, opened 2026-08-13
[TAO-0][Bugfix] NVBug 6597811: Use the canonical public HTTPS install URL

ramanathan831:
## Reported issue

NVBug 6597811: README installation used GitHub SSH for a public repository and the obsolete `tao-skills-bank` slug. Fresh machines without an authorized GitHub SSH key failed before the plugin was installed.

The README's Claude and Codex marketplace commands used `git@github.com:...`. QA reproduced `Could not read from remote repository` on machines without GitHub SSH setup; HTTPS succeeded.

## Root cause

Public install documentation assumed developer SSH credentials and relied on a redirect from an obsolete repository name.

## Implementation

Use `https://github.com/NVIDIA-TAO/tao-skill-bank.git` in both marketplace commands and update the heading/raw installer links to the canonical slug.

- No `git@github.com` or `NVIDIA-TAO/tao-skills-bank` remains in README.
- `git ls-remote https://github.com/NVIDIA-TAO/tao-skill-bank.git HEAD` succeeded without SSH.
- `python -m pytest -q scripts/tests/test_install_codex_agents.py` — **3 passed**

## Residual risk

Revalidation avoided mutating the user's permanent marketplace configuration; it verified the exact public transport and repository target. Other non-README stale slug consumers are handled separately by the audit PR.

The documented install entrypoint now names the canonical public repository over HTTPS, matching the actual default marketplace source and working on hosts without SSH credentials.

The change does not rely on repository redirects, require SSH keys for a public checkout, or add a fallback that would leave the primary command broken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# NVIDIA [TAO Skill Bank](https://github.com/NVIDIA-TAO/tao-skills-bank)
+# NVIDIA [TAO Skill Bank](https://github.com/NVIDIA-TAO/tao-skill-bank)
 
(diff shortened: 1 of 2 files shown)


ramanathan831:
Closing because the reported install failure was already fixed by merged PR #149, independently of PR #198. The NVBug 6597811 reproduction no longer finds a git@github.com URL or the obsolete tao-skills-bank slug in README.md. Keeping this older patch would conflict with the newer release-tagged installation strategy.

PR #134, opened 2026-08-13
[TAO-0][Bugfix] NVBug 6598785: Reconcile a completed container after wrapper interruption

ramanathan831:
## Reported issue

NVBug 6598785: interrupting the host wrapper after a container completed left the stage marked `running`; the completed output could not be committed safely and rerunning could consume the final attempt or relaunch completed work.

## Original reproduction

1. Launch an IAA container stage with its deterministic container name and fresh-output contract.
2. Let the auto-removed container finish successfully and write its output and `Execution status: PASS` log marker.
3. Terminate the host wrapper before it atomically records the terminal command status.
4. Rerun the same user-facing stage command.

Before this change, the status remained `running` even though Docker no longer had the container. The wrapper could not distinguish completed work from an abandoned attempt and did not recover the stage.

## Root cause

Terminal success existed only in wrapper memory. The persisted running record had enough launch identity to locate the prior work, but the resume path did not reconcile that record against Docker state, the TAO log, and output freshness.

## Implementation summary

- Validate the complete persisted command identity before recovery.
- Inspect a retained container exit code when available.
- For an auto-removed container, require the final TAO log status to be `PASS`.
- In both cases, require every declared output to be nonempty, non-symlinked, and newer than the recorded attempt start.
- Atomically close the existing attempt as successful and return without relaunching.
- Keep ambiguous or failed evidence fail-closed.

## Regression coverage

Added interrupted-wrapper tests covering successful reconciliation without a second launch and rejection of a PASS log when a required fresh output is missing. Adjacent container identity, GPU-scope, command-status, and commit behavior remained covered.

## Exact post-fix verification

The original on-disk `running` status + PASS log + fresh output topology reconciled attempt 1 to `ok`, preserved the output, and did not call Docker launch. Removing the fresh output made reconciliation fail as intended.

- `python -m pytest -q scripts/tests` — **260 passed, 2 skipped**
- focused interruption and adjacent IAA container/commit tests — **passed**

Recovery is implemented at the durable command-status boundary where the lost wrapper transition occurred. It reconstructs the missing transition only when independent launch identity, backend/log completion, and artifact evidence agree, so every caller gets the same crash-safe behavior.

The implementation does not delete the running status, blindly trust a PASS string, assume a missing container succeeded, or relaunch work just to make the stage usable. It also does not accept outputs from an older attempt.

## Residual risks or limitations

An auto-removed container with no durable PASS marker cannot be proven successful and is intentionally not reconciled. Such ambiguous state still requires the normal retry/manual recovery path rather than guessing.

### DIFF
--- a/skills/applications/tao-run-deft-iaa/scripts/run_deft_container.py
+++ b/skills/applications/tao-run-deft-iaa/scripts/run_deft_container.py
@@ -254,6 +254,147 @@ def _container_is_running(container_name: str) -> bool:
     )
 
 
+def _container_exit_code(container_name: str) -> int | None:
+    """Return a retained container's exit code, or None after Docker --rm."""
+    inspected = subprocess.run(
+        [
+            "docker",
+            "container",
+            "inspect",
+            "--format",
+            "{{.State.ExitCode}}",
+            container_name,
+        ],
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+    if inspected.returncode == 0:
+        try:
+            return int(inspected.stdout.strip())
+        except ValueError as exc:
+            raise ValueError(
+                f"Docker returned an invalid exit code for {container_name}: "
+                f"{inspected.stdout.strip()!r}"
+            ) from exc
+    combined = (inspected.stdout + inspected.stderr).lower()
+    if "no such object" in combined or "no such container" in combined:
+        return None
+    raise ValueError(
+        f"cannot inspect prior container {container_name}; Docker said: "
+        f"{(inspected.stderr or inspected.stdout).strip() or 'unknown error'}"
+    )
+
+
+def _last_execution_status(log_path: pathlib.Path) -> str | None:
+    """Read only the bounded log tail and return its final TAO status marker."""
+    with log_path.open("rb") as handle:
+        size = log_path.stat().st_size
+        handle.seek(max(0, size - 64 * 1024))
+        tail = handle.read().decode("utf-8", errors="replace")
+    statuses = re.findall(
+        r"(?m)^Execution status:\s*(PASS|FAIL)\s*$", tail, flags=re.IGNORECASE
+    )
+    return statuses[-1].upper() if statuses else None
+
+
+def _validate_interrupted_identity(
+    existing: dict[str, Any],
+    *,
+    name: str,
+    image_kind: str,
+    image: str,
+    command: list[str],
+    passed_hf_token: bool,
+    container_name: str,
+    cidfile_path: pathlib.Path,
+    log_path: pathlib.Path,
+    fresh_outputs: list[str],
+) -> None:
+    expected = {
+        "schema_version": "1",
+        "workflow": "tao-run-deft-iaa",
+        "kind": "container",
+        "name": name,
+        "image_kind": image_kind,
+        "image": image,
+        "command": command,
+        "command_sha256": command_sha256(command),
+        "passed_hf_token": passed_hf_token,
+        "container_name": container_name,
+        "cidfile": str(cidfile_path),
+        "log_path": str(log_path),
+        "fresh_outputs": fresh_outputs,
+    }
+    mismatches = [key for key, value in expected.items() if existing.get(key) != value]
+    if mismatches:
+        raise ValueError(
+            "existing running status does not match the approved command identity: "
+            + ", ".join(mismatches)
+        )
+
+
+def _reconcile_interrupted_success(
+    existing: dict[str, Any],
+    *,
+    status_path: pathlib.Path,
+    log_path: pathlib.Path,
+    fresh_outputs: list[str],
+    container_name: str,
+) -> bool:
+    """Close a status left running only when durable success evidence agrees."""
+    started_ns = existing.get("started_ns")
+    if (
+        not isinstance(started_ns, int)
+        or isinstance(started_ns, bool)
+        or started_ns < 1
+        or not log_path.is_file()
+        or log_path.stat().st_size == 0
+        or log_path.is_symlink()
+        or log_path.resolve() != log_path
+    ):
+        return False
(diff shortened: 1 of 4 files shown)


skills/applications/tao-run-deft-iaa/scripts/run_deft_container.py
```diff
@@ -254,6 +254,147 @@ def _container_is_running(container_name: str) -> bool:
...
+        f"cannot inspect prior container {container_name}; Docker said: "
+        f"{(inspected.stderr or inspected.stdout).strip() or 'unknown error'}"
+    )
+
+
+def _last_execution_status(log_path: pathlib.Path) -> str | None:
+    """Read only the bounded log tail and return its final TAO status marker."""
+    with log_path.open("rb") as handle:
+        size = log_path.stat().st_size
+        handle.seek(max(0, size - 64 * 1024))
```
vpraveen-nv: **A 64 KiB tail can miss `Execution status: PASS`, and the failure mode is the bug re-appearing.**

When the marker falls outside the tail, `_last_execution_status` returns `None`. If the container was `--rm`'d so `_container_exit_code` is also `None`, `_reconcile_interrupted_success` returns `False` and the run lands back on exactly the terminal state NVBug 6598785 describes — with the same 45 minutes of GPU work stranded.

The bug's own evidence shows `pool_embed.log` ending with a tqdm bar immediately before the marker, so today's logs are comfortably inside 64 KiB. It's the stage that emits a teardown dump or a stack of warnings after the marker that would regress silently. Scanning backwards in chunks until the first marker is found, or logging "no status marker within the last 64 KiB" when the tail is exhausted, would make the miss visible instead of indistinguishable from a genuine failure.

skills/applications/tao-run-deft-iaa/scripts/run_deft_container.py
```diff
@@ -254,6 +254,147 @@ def _container_is_running(container_name: str) -> bool:
...
+    finished_at = dt.datetime.fromtimestamp(
+        log_path.stat().st_mtime, tz=dt.timezone.utc
+    ).isoformat(timespec="seconds")
+    _atomic_json(
+        status_path,
+        {
+            **existing,
+            "finished_at": finished_at,
+            "status": "ok",
+            "exit_code": 0,
```
vpraveen-nv, line 394: **`exit_code` is hardcoded 0 while `docker_exit_code` records provenance — worth a note in the status schema.**

In the `container_log` branch Docker is gone, so `docker_exit_code` is synthesised as 0 too and only `reconciliation_source` distinguishes "Docker told us it exited 0" from "we inferred success from a log line".

That's a defensible design and the surrounding guards (fresh outputs newer than `started_ns`, non-empty non-symlink log, identity match) are what make it safe. Since `reconciled_after_wrapper_exit` is the flag an auditor would filter on, it's worth making sure `audit_deft_run.py` surfaces it in the run report — a reconciled stage should be visibly distinct from one the wrapper closed itself.

skills/applications/tao-run-deft-iaa/scripts/run_deft_container.py
ramanathan831: Fixed in cff8580. `_last_execution_status` now streams the complete log with bounded memory and retains the final marker, eliminating the arbitrary 64 KiB evidence window. The end-to-end interrupted-wrapper reproduction now places PASS before more than 64 KiB of trailing diagnostics, and a second regression proves a later FAIL overrides an earlier PASS. Container evidence tests pass 8/8.

skills/applications/tao-run-deft-iaa/scripts/run_deft_container.py
ramanathan831: Documented in cff8580. `exit_code` is the wrapper contract outcome after the complete evidence set establishes success. `docker_exit_code=0` may be directly inspected or inferred for an auto-removed container; `reconciliation_source` (`docker` vs `container_log`) is the authoritative provenance discriminator, with `reconciled_after_wrapper_exit=true` in both cases. This preserves the existing downstream zero-exit invariant without hiding how it was established.

christinayyw:

**Latest run:** [tao-sb-OSS-develop #348](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/348/) — `ee8c9a9c` @ 2026-08-27 23:20 UTC

|---|---|---|---|---|
| `skills/applications/tao-run-deft-iaa` | — | `claude` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/348/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-claude/skill-eval-report.html)<br><details><summary>Previous Reports</summary>[Build #186 Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/186/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-claude/skill-eval-report.html)</details> |
| `skills/applications/tao-run-deft-iaa` | — | `codex` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/348/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-codex/skill-eval-report.html)<br><details><summary>Previous Reports</summary>[Build #186 Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/186/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-codex/skill-eval-report.html)</details> |


vpraveen-nv (approved):

PR #138, opened 2026-08-13
[TAO-0][Bugfix] NVBug 6603466: Validate checkpoints against the train-attempt lineage

ramanathan831:
## Reported issue

NVBug 6603466: retrying a train stage could make the run terminal because a valid checkpoint produced by the first attempt was considered older than the retry attempt.

## Original reproduction

1. Start a train stage and produce a valid epoch checkpoint.
2. Let the wrapper fail after checkpoint creation, then rerun the same stage as attempt 2.
3. Publish and commit the best checkpoint.

The checkpoint timestamp was newer than attempt 1 but older than attempt 2. Before this change, publication/commit compared it only with attempt 2's start and rejected it as stale, although both attempts belonged to the same train-stage lineage.

## Root cause

Each wrapper retry overwrote the only persisted start timestamp used for checkpoint freshness. The system modeled attempt freshness but not stage-lineage freshness, so downstream publication could not distinguish a valid earlier-attempt artifact from output predating the entire stage.

## Implementation summary

- Persist `lineage_started_ns` on the first attempt and carry it through retries.
- Validate lineage timestamps on every consumer.
- Select and publish checkpoints only when they are nonempty and no older than the lineage.
- Record whether best-checkpoint selection used metric evidence or the explicit newest-checkpoint fallback.
- Audit and commit against the same lineage boundary.

## Regression coverage

Added tests for a retry publishing an earlier-attempt checkpoint in the same lineage and for rejecting a checkpoint older than the lineage before creating a best-checkpoint link or metadata.

## Exact post-fix verification

The exact timestamp topology now publishes the earlier-attempt checkpoint, records `newest_fallback` with `metric: null`, and validates its provenance. Moving the checkpoint one nanosecond before the lineage start raises before any publication artifact is created.

- `python -m pytest -q scripts/tests` — **260 passed, 2 skipped**
- focused checkpoint tests — **2 passed**
- adjacent IAA checkpoint/commit/audit tests — **passed**

The lifecycle model now has a durable lineage timestamp, and every producer/consumer of checkpoint provenance uses it consistently. Retries no longer redefine when the logical train stage began, while artifacts from before that stage remain invalid.

The fix does not disable freshness checks, touch checkpoint mtimes, blindly accept any preexisting checkpoint, or special-case attempt 2. It also validates before replacing the canonical best-checkpoint publication.

## Residual risks or limitations

Freshness still relies on filesystem nanosecond timestamps, as the existing artifact contract does. The lineage closes the retry bug without changing that broader provenance model.

### DIFF
--- a/skills/applications/tao-run-deft-iaa/scripts/checkpoint_contract.py
+++ b/skills/applications/tao-run-deft-iaa/scripts/checkpoint_contract.py
@@ -5,6 +5,7 @@
 
 from __future__ import annotations
 
+import datetime as dt
 import json
 import os
 import pathlib
@@ -13,6 +14,62 @@ from typing import Any
 
 BEST_RELPATH = pathlib.Path("best/clip_best_val_t2i_mAP.pth")
 METADATA_RELPATH = pathlib.Path("best/clip_best_val_t2i_mAP.json")
+
+
+def checkpoint_lineage_started_ns(
+    status: dict[str, Any], run_started_at: str
+) -> int:
+    """Resolve a retry lineage lower bound anchored to the immutable run.
+
+    New statuses carry ``lineage_started_ns`` explicitly. Legacy attempt-1
+    statuses use their own start, while legacy retries—whose first-attempt
+    timestamp was never persisted—fall back to the run start rather than the
+    retry start that caused NVBug 6603466. The run anchor also prevents a
+    status copied from an earlier occupant of the results path from widening
+    the checkpoint window before this run existed.
+    """
+    started_ns = status.get("started_ns")
+    attempt = status.get("attempt", 1)
+    if (
+        not isinstance(started_ns, int)
+        or isinstance(started_ns, bool)
+        or started_ns < 1
+    ):
+        raise ValueError("train command status started_ns must be a positive integer")
+    if (
+        not isinstance(attempt, int)
+        or isinstance(attempt, bool)
+        or not 1 <= attempt <= 2
+    ):
+        raise ValueError("train command status attempt must be 1 or 2")
+    if not isinstance(run_started_at, str) or not run_started_at.strip():
+        raise ValueError("run started_at must be a non-empty timestamp")
+    try:
+        run_started = dt.datetime.fromisoformat(run_started_at.replace("Z", "+00:00"))
+    except ValueError as exc:
+        raise ValueError("run started_at must be an ISO-8601 timestamp") from exc
+    if run_started.tzinfo is None:
+        raise ValueError("run started_at must include a timezone")
+    run_started_ns = int(run_started.timestamp() * 1_000_000_000)
+    if run_started_ns < 1 or run_started_ns > started_ns:
+        raise ValueError("run started_at must be no later than the train attempt")
+
+    if "lineage_started_ns" in status:
+        lineage_started_ns = status["lineage_started_ns"]
+    elif attempt == 1:
+        lineage_started_ns = started_ns
+    else:
+        lineage_started_ns = run_started_ns
+    if (
+        not isinstance(lineage_started_ns, int)
+        or isinstance(lineage_started_ns, bool)
+        or not run_started_ns <= lineage_started_ns <= started_ns
+    ):
+        raise ValueError(
+            "train command status lineage_started_ns must fall between the "
+            "immutable run start and current attempt start"
+        )
+    return lineage_started_ns
 
 
 def _absolute_lexical(path: pathlib.Path | str, name: str) -> pathlib.Path:
--- a/skills/applications/tao-run-deft-iaa/scripts/checkpoint_contract.py
+++ b/skills/applications/tao-run-deft-iaa/scripts/checkpoint_contract.py
@@ -93,7 +93,7 @@ def validate_best_checkpoint(
     )
     if source.stat().st_mtime_ns < started_ns:
         raise ValueError(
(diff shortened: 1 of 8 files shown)


skills/applications/tao-run-deft-iaa/scripts/run_iaa_stage.py
```diff
@@ -544,6 +544,16 @@ def publish_checkpoint(args: argparse.Namespace) -> dict[str, Any]:
     started_ns = payload.get("started_ns")
     if not isinstance(started_ns, int) or isinstance(started_ns, bool) or started_ns < 1:
         raise ValueError("train command status started_ns must be a positive integer")
+    lineage_started_ns = payload.get("lineage_started_ns", started_ns)
```
vpraveen-nv: **The default silently reproduces the bug for every run that predates this merge.**

When a status payload has no `lineage_started_ns`, this falls back to `started_ns` — the retry's start — which is exactly the comparison NVBug 6603466 says is wrong. Any run whose `train.status.json` was written by the current code has no such key, so the stale-checkpoint `ValueError` still fires and the run still dies one stage from the finish.

That population isn't hypothetical: the bug reports two occurrences in ordinary runs on two machines, and its precondition is a wrapper interruption, which is also what PR #134 (NVBug 6598785) is about. If the intent is that in-flight runs are not recoverable and must be restarted, that's worth stating; if they should recover, the fallback needs to derive the lineage start from the earliest attempt on disk rather than defaulting to the current one.

skills/applications/tao-run-deft-iaa/scripts/checkpoint_contract.py
```diff
@@ -93,14 +93,21 @@ def validate_best_checkpoint(
     )
     if source.stat().st_mtime_ns < started_ns:
         raise ValueError(
-            f"selected checkpoint predates the current train attempt: {source}"
+            f"selected checkpoint predates the train attempt lineage: {source}"
```
vpraveen-nv, line 153: **What still bounds the checkpoint to *this* run once the window widens?**

Comparing against the lineage start instead of the attempt start is the correct fix for the reported failure. It does mean the accepted window now reaches back to the first attempt, so a checkpoint left in `train_dir` by an earlier, unrelated occupant of the same results directory would pass this check where it previously failed.

`get_current_checkpoint(..., earliest_mtime_ns=lineage_started_ns)` plus the `train_dir` scoping probably makes that unreachable in practice, since the run directory is immutable per run. Worth confirming that's the reasoning, ideally as a comment here — this line is the one a future reader will question.

vpraveen-nv (approved):

skills/applications/tao-run-deft-iaa/scripts/run_iaa_stage.py
ramanathan831: Fixed in 476a03b. All publish/commit/audit consumers now call one resolver. A legacy attempt-1 status uses its own start; a legacy attempt-2 status, where the original timestamp was never persisted, uses immutable `deft_state.started_at` rather than the retry start. The regression publishes a checkpoint created after run start but before the legacy retry, proving the original failure no longer occurs.

skills/applications/tao-run-deft-iaa/scripts/checkpoint_contract.py
ramanathan831: Fixed in 476a03b. The widened attempt lineage is now constrained by the immutable run start: explicit `lineage_started_ns` must fall between `deft_state.started_at` and the current attempt start, and legacy retry fallback is exactly that run boundary. A status/checkpoint lineage copied from an earlier occupant therefore fails before checkpoint validation. Documentation and regression coverage record this invariant; checkpoint/container tests pass 9/9.

christinayyw:

**Latest run:** [tao-sb-OSS-develop #343](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/343/) — `e99c8684` @ 2026-08-27 21:03 UTC

|---|---|---|---|---|
| `skills/applications/tao-run-deft-iaa` | — | `claude` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/343/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-claude/skill-eval-report.html)<br><details><summary>Previous Reports</summary>[Build #190 Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/190/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-claude/skill-eval-report.html)</details> |
| `skills/applications/tao-run-deft-iaa` | — | `codex` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/343/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-codex/skill-eval-report.html)<br><details><summary>Previous Reports</summary>[Build #190 Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/190/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-iaa-codex/skill-eval-report.html)</details> |

PR #141, opened 2026-08-14
[TAO-0][Bugfix] Render platform action contracts safely on Kubernetes

ramanathan831:
## Issues

DEFT action requests need multiple producer-defined mounts, absolute-path aliases, safe argv, output coverage, and narrowly scoped credential forwarding. The fixed Kubernetes template collapsed these requirements into one PVC target. Review also reproduced two fail-open behaviors: a whole Secret was imported even when only one or no environment key was approved, and unsupported action-request schema versions were accepted.

Arif's review identified another contract gap: the renderer rejected every `spec_bundle.mode=config` request. That excluded DEFT AOI train/evaluate/inference/RCA and mining actions, plus Cosmos-RL train, even though config mode is a first-class producer/consumer contract.

## Original reproductions and observed failures

1. Prepare an IAA action request containing the run, config, patches, cache, and (for image actions) dataset mounts. The previous template cannot represent the producer's distinct targets/aliases.
2. Set `forward_env=[]` with a credential Secret, or `forward_env=["HF_TOKEN"]`. The previous template imports the whole Secret through `envFrom`.
3. Change `schema_version` to `999`. The previous renderer still renders it.
4. Change a valid action request to `mode=config`, supply a nested `spec`, `config_format=yaml`, and a command containing `{config_path}`. The renderer fails before launch with `Kubernetes action rendering requires spec_bundle.mode=args`.

## Root causes

- The Kubernetes consumer owned a fixed application-specific mount shape instead of rendering the producer's declared action contract.
- Credential rendering treated Secret existence as authorization for every key rather than projecting the approved allowlist.
- Only the staging-map shape was validated; the request envelope version was trusted.
- The consumer implemented only the args branch of the mode-discriminated spec-bundle contract. It had no owned step to serialize config content, stage the resulting file, mount it into the pod, or substitute its compute-frame path.

## Implementation

- Add a generic action-job renderer and template driven by the validated action request plus an explicit source-to-PVC staging map.
- Render every distinct mount and alias with the correct read-only flags; validate output coverage, command/argv, resource shape, paths, and Kubernetes names.
- Normalize invalid job IDs while preserving the original ID as an annotation.
- Project each approved credential key through individual `env[].valueFrom.secretKeyRef` entries; never use whole-Secret `envFrom`, and reject a Secret when the allowlist is empty.
- Fail closed on missing or unsupported request schema versions.
- Add `materialize-config` for `mode=config` bundles. It canonically serializes YAML/JSON/TOML specs into atomic, SHA-256-addressed, owner-only files.
- Require the exact generated file in the staging receipt, verify its bytes again at render time, reject symlinks/stale content, mount the staged file read-only at a content-addressed pod path, and substitute that path for `{config_path}`.
- Preserve simple commands as native Kubernetes argv. Preserve producer-owned multi-line config shell commands (including Cosmos-RL hook discovery and heredocs) verbatim under `/bin/sh -c`; config values are never interpolated as shell text.
- Add contract documentation and template validation coverage.

## Why this is a root-cause fix

The consumer now implements both branches of the producer/consumer abstraction. Config content stays producer-owned while serialization, placement, and compute-frame path substitution stay platform-owned. The rendered pod consumes the same bytes that were derived from the approved bundle, instead of relying on an unrelated path or mutable hand-authored file.

We intentionally avoided the shortcut of re-emitting config requests as `mode=args`, adding DEFT-stage special cases, trusting an arbitrary `--config-source`, or copying config values into shell text. Those approaches would mask the missing consumer behavior, weaken provenance, or introduce a second platform-specific action contract. The mount renderer remains application-neutral.

Credential authorization is enforced at key granularity, and version compatibility is enforced at the request boundary. We also avoided hardcoded mount counts, application-path conditionals, whole-Secret fallbacks, and permissive future-schema handling.

## Regression coverage and post-fix verification

- Focused Kubernetes action-renderer suite: **43 passed**.
- Kubernetes renderer plus adjacent single-pod and multi-node template suites: **72 passed**.
- Full `scripts/tests`: **302 passed, 2 skipped**.
- `scripts/validate-skills.sh`: **passed**.
- `scripts/stamp_versions.py --check`: **passed**.
- `git diff --check`: **passed**.
- The original config-mode CLI reproduction now materializes the spec, includes the generated file in the staging receipt, renders the Job, removes the literal `{config_path}`, and produces a verified read-only file mount.
- Variations cover YAML, JSON, nested TOML and inline-table arrays, TOML null rejection, atomic/idempotent materialization, full content hashes, altered/stale content, symlinks, missing staging entries, args/config mode mismatch, commands missing `{config_path}`, explicit args-mode shells, and Cosmos-RL-style multi-line shell commands.
- Existing IAA mount, output, naming, argv, environment, and Secret-projection regressions continue to pass.

## Residual risk

Live API-server validation remains unavailable in this environment because it has no Kubernetes client/cluster context, namespace, PVC, or S3 target. The manifest is parsed in regression tests, but `kubectl apply --dry-run=server`/`kubeconform` was skipped. No live Kubernetes execution claim is made. TOML has no null value, so a TOML config containing `null` fails closed rather than silently dropping or changing that value.

### DIFF
--- a/scripts/ci/render_and_validate_templates.py
+++ b/scripts/ci/render_and_validate_templates.py
@@ -58,6 +58,26 @@ K8S_VALS = {
     "MOUNT_PATH": "/data",
     "SHM_SIZE": "16Gi",
     "PVC_CLAIM": "edgeai-datasets",
+    # Platform-neutral action request template. JSON is valid YAML and is the
+    # exact encoding used by render_action_job.py.
+    "JOB_NAME_JSON": '"nightly-contract-0001"',
+    "JOB_ID_JSON": '"nightly-contract-0001"',
+    "NAMESPACE_JSON": '"default"',
+    "TTL_SECONDS_JSON": "3600",
+    "IMAGE_PULL_SECRETS_JSON": '[{"name":"ngc-pull-secret"}]',
+    "IMAGE_JSON": '"nvcr.io/nvidia/tao/tao-toolkit:7.1.0-pyt"',
+    "COMMAND_JSON": '["python3"]',
+    "ARGS_JSON": '["-c","print(1)"]',
+    "NUM_GPUS_JSON": '"1"',
+    "ENV_FROM_JSON": '[]',
+    "ENV_JSON": '[{"name":"HOME","value":"/tmp"}]',
+    "VOLUME_MOUNTS_JSON": (
+        '[{"name":"dshm","mountPath":"/dev/shm","readOnly":false},'
+        '{"name":"workspace","mountPath":"/results",'
+        '"subPath":"jobs/nightly/results","readOnly":false}]'
+    ),
+    "SHM_SIZE_JSON": '"16Gi"',
+    "PVC_CLAIM_JSON": '"edgeai-datasets"',
 }
 SLURM_VALS = {
     **COMMON,
--- a/scripts/ci/render_and_validate_templates.py
+++ b/scripts/ci/render_and_validate_templates.py
@@ -69,7 +69,6 @@ K8S_VALS = {
     "COMMAND_JSON": '["python3"]',
     "ARGS_JSON": '["-c","print(1)"]',
     "NUM_GPUS_JSON": '"1"',
-    "ENV_FROM_JSON": '[]',
     "ENV_JSON": '[{"name":"HOME","value":"/tmp"}]',
     "VOLUME_MOUNTS_JSON": (
         '[{"name":"dshm","mountPath":"/dev/shm","readOnly":false},'
--- a/scripts/tests/test_k8s_action_renderer.py
+++ b/scripts/tests/test_k8s_action_renderer.py
@@ -130,7 +130,7 @@ def test_no_credentials_or_registry_secret_means_no_dangling_secret_refs():
     job, action = container(render())
     pod = job["spec"]["template"]["spec"]
     assert pod["imagePullSecrets"] == []
-    assert action["envFrom"] == []
+    assert "envFrom" not in action
 
 
 def test_forwarded_credentials_are_secret_references_not_inline_values():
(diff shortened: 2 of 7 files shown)


arif-ahmed-nv:
`render_action_job.py` rejects `spec_bundle.mode != "args"`, but most DEFT stages are `mode=config` — train/evaluate/inference/rca and all three mining stages in DEFT AOI, and the cosmos-rl train action besides. As written, none of those can route through this renderer.

On our side (#186) the consumer materializes the spec into a rendered file and substitutes its path into the command's `{config_path}` placeholder — with your staging map, that looks like it could be one more staged file plus a volumeMount, reusing the same subPath validation you already have.

If config-mode is deliberately out of scope for this PR, could the `RenderError` say what the caller should do instead of only refusing? Something like "materialize the spec and re-emit as mode=args" would keep a first-time consumer from reading it as "config-mode workflows cannot run on Kubernetes."

Happy to pick this up as a follow-up once this lands — we're planning to converge the two k8s renderers after the merge anyway.

ramanathan831:
Addressed the config-mode gap from your review in 1881a58. The Kubernetes consumer now materializes the nested spec canonically, requires that exact generated file in the staging receipt, verifies its bytes, mounts it read-only at a content-addressed pod path, and substitutes `{config_path}`. This covers YAML/JSON/TOML, native commands, and Cosmos-RL-style multi-line shell commands. I intentionally did not re-emit config bundles as mode=args because that would bypass the producer/consumer contract and weaken config provenance. Focused renderer: 43 passed; full scripts/tests: 302 passed, 2 skipped; validate-skills passed.

arif-ahmed-nv (approved):

christinayyw:

**Latest run:** [tao-sb-OSS-develop #360](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/360/) — `1881a58f` @ 2026-08-28 03:42 UTC

|---|---|---|---|---|
| `skills/platform/tao-run-on-kubernetes` | — | `claude` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/360/artifact/jenkins-eval-reports/skills-platform-tao-run-on-kubernetes-claude/skill-eval-report.html)<br><details><summary>Previous Reports</summary>[Build #320 Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/320/artifact/jenkins-eval-reports/skills-platform-tao-run-on-kubernetes-claude/skill-eval-report.html)<br>[Build #192 Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/192/artifact/jenkins-eval-reports/skills-platform-tao-run-on-kubernetes-claude/skill-eval-report.html)</details> |
| `skills/platform/tao-run-on-kubernetes` | — | `codex` | ✅ Pass (⚠️ w/ self-correction) | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/360/artifact/jenkins-eval-reports/skills-platform-tao-run-on-kubernetes-codex/skill-eval-report.html)<br><details><summary>Previous Reports</summary>[Build #320 Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/320/artifact/jenkins-eval-reports/skills-platform-tao-run-on-kubernetes-codex/skill-eval-report.html)<br>[Build #192 Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/192/artifact/jenkins-eval-reports/skills-platform-tao-run-on-kubernetes-codex/skill-eval-report.html)</details> |

PR #139, opened 2026-08-13
[TAO-0][Bugfix] Audit: Enforce strict Data Services specs and canonical repository URLs

ramanathan831:
## Discovered issues

The related-pattern audit confirmed two previously unaddressed production defects:

1. Five Data Services skill validators silently coerced invalid YAML scalar types (for example `true` as a count, `1.5` as an integer count, numeric model paths, and string booleans). Some non-finite numeric values also passed range checks.
2. Three non-README install/routing consumers and the NemoClaw setup script still referenced the redirected `NVIDIA-AI-Blueprints/tao-skill-bank` slug after the canonical public URL had moved to `NVIDIA-TAO/tao-skill-bank`.

## Original reproductions

On latest `main`, invoke each validator through a fresh process with otherwise valid specs and mutate one field:

- nearest neighbors: `topn: true`, fractional counts, boolean distance threshold;
- image embeddings: boolean/fractional batch size or numeric model paths;
- OD gap analysis: boolean/non-finite thresholds and numeric KPI;
- detection KPI: fractional recall points, string booleans, boolean thresholds;
- unique-neighbor matching: boolean/fractional counts and non-boolean flags.

All five validators printed/returned success for at least one invalid value because Python conversion made it look valid. Separately, repository search found executable/card consumers of the old redirected slug.

## Root causes

- Validator code normalized with `int(...)`, `float(...)`, `str(...)`, and truthiness before enforcing YAML types. Python booleans are integers, fractional values can truncate, strings can masquerade as paths/flags, and `NaN` defeats ordinary range comparisons.
- The earlier URL fix covered the README path but not all install and routing consumers, leaving copy-pasted stale repository identifiers.

## Implementation summary

- Enforce native YAML types, integer-vs-boolean distinctions, finite numeric constraints, ranges, enums, nonempty strings, mappings, lists, and booleans before normalization in all five affected validators.
- Preserve each skill's documented valid defaults and output behavior.
- Update all confirmed stale non-README repository consumers to the canonical HTTPS URL.
- Add parameterized regression coverage for every validator and a repository-wide stale-URL guard.

## Regression coverage

- `scripts/tests/test_data_service_spec_validation.py` covers valid specs plus wrong types, fractional counts, booleans-as-numbers, non-finite values, and related optional fields across all five skills.
- `scripts/tests/test_repository_urls.py` prevents the redirected slug from returning outside intentionally preserved historical text.

## Exact post-fix verification

[...]

PR #142, opened 2026-08-14
[TAO-0][Docs] Use canonical TAO Skill Bank checkout name

xiaosean:
## Summary

- Replace obsolete `tao-skills-external` local checkout examples in the README with `tao-skill-bank`.
- Align the repository structure tree with the canonical checkout name.
- Remove README from the temporary stale-name test exception so the canonical name remains enforced.
- Keep the supported `TAO_SKILL_BANK_MARKETPLACE` override behavior unchanged.

## Related

- Complements #139, which converted every other stale-name consumer. After this PR, the only repository-wide stale-name match is the self-excluded `STALE_NAMES` test tuple; therefore this closes NVBug 6589974 even though the bug report’s original occurrence count is now stale.
- Complements #133, which handles stale public repository URLs and public HTTPS install commands.

## Validation

- `python -m pytest -q scripts/tests/test_repository_urls.py` — 1 passed
- `bash scripts/validate-skills.sh` — passed
- `git diff --check`


README.md
```diff
@@ -36,8 +36,8 @@ Override the source with `TAO_SKILL_BANK_MARKETPLACE=…` and
 `TAO_SKILL_BANK_REF=…` to use a fork, pinned ref, or local absolute path:
 
 ```bash
-cd /absolute/path/to/tao-skills-external
-TAO_SKILL_BANK_MARKETPLACE=/absolute/path/to/tao-skills-external \
+cd /absolute/path/to/tao-skill-bank
```
vpraveen-nv, line 74: This PR should also drop the `README.md` entry from `scripts/tests/test_repository_urls.py` (added by #139), whose comment literally says *"PR #142 owns README's canonical checkout-name migration. Keep the exception explicit..."*. That exception was a temporary hand-off to this PR; once these four lines land it is stale, and `README.md` becomes the one file in the repo where `tao-skills-external` / `tao-skills-bank` can silently reappear — the same regression NVBug 6589974 was filed for.

I applied this diff onto current `main`, changed `EXCEPTIONS = {Path("README.md")}` to `EXCEPTIONS = set()`, and `test_consumers_use_canonical_tao_skill_bank_name` passes — so the exception can be deleted in this PR with no other changes. Please fold that one-line removal (plus the now-obsolete comment) into this diff so the fix is guarded rather than just applied.

README.md
```diff
@@ -57,7 +57,7 @@ This installs the bundle to `~/.codex/plugins/cache/tao-local-plugins/tao-skill-
 For a local zip or clone, use the absolute path instead of the Git URL:
 
 ```bash
-codex plugin marketplace add /absolute/path/to/tao-skills-external
+codex plugin marketplace add /absolute/path/to/tao-skill-bank
```
vpraveen-nv, line 95: Scope note for whoever merges, not a change request on these lines: the bug's "66 occurrences across 32 files" figure is stale. #139 already converted every other consumer, including the `TAO_SKILL_BANK_PATH` fallback defaults, which now read `Path.home() / "tao-skill-bank"` in `resolve_tao_image.py`, `resolve_tao_model.py`, `list_tao_models.py`, `list_tao_capabilities.py` and `check_tao_launch_preflight.py`.

After this PR the only remaining repo-wide match is the `STALE_NAMES` tuple inside the test itself, which is self-excluded — so this PR does close NVBug 6589974. Worth stating in the PR body so the bug isn't closed against a count that no longer matches reality. Separately, the branch is 48 commits behind `main`; a rebase before merge would avoid a surprising 3-way resolution.

vpraveen-nv (approved):
minor nitpicks

README.md
ramanathan831: Addressed in 6ed808d: the temporary README exception and its hand-off comment are removed. The repository URL regression test now scans README and passes against the canonicalized examples.

README.md
ramanathan831: Updated the PR body to record that #139 converted the other consumers, the original occurrence count is stale, and only the self-excluded test tuple remains. I also merged current main in 6ed808d, eliminating the branch lag.

PR #149, opened 2026-08-18
[TAO-000][Docs] Point install instructions at the latest release build  (7.1.0)

christinayyw:
The Claude Code and Codex install commands pulled the marketplace from the default branch, so a fresh install picked up whatever `main` held at that moment rather than the published release build. Pin every install path to the latest release tag (7.1.0)


vpraveen-nv (approved):
LGTM.

PR #150, opened 2026-08-19
[TAO-2230][Bugfix] deft-od: give the containers an identity alongside --user

arihantj-nv:
Fixes NVBug 6603226 and NVBug 6635686. They are the same defect reported twice, from two hosts (uid 159080 and uid 2583), so this is one PR.

## The defect

Every documented launch passes `--user "$(id -u):$(id -g)"` and nothing else. A uid with no `/etc/passwd` entry inside the image — every uid but 1000 on the TAO images — has no name, so `getpass.getuser()` falls through to `pwd.getpwuid()` and raises. torch calls it while setting up its inductor cache directory, so the container dies at import, before the requested action starts:

```
KeyError: 'getpwuid(): uid not found: 159080'
  File ".../torch/_inductor/runtime/cache_dir_utils.py", line 23, in default_cache_dir
    sanitized_username = re.sub(pattern, "_", getpass.getuser())
```

`HOME` is unset for the same reason and defaults to `/`, which is not writable, so matplotlib and cupy fail on their cache directories.

This is invisible on a uid-1000 workstation and fires on every shared or cloud host, which is why it survived the runs this skill was developed against.

## The fix

Pre-flight exports `DOCKER_IDENTITY` beside `EXTRA_MOUNTS`, and all 11 launches that pass `--user` carry it:

```bash
DOCKER_IDENTITY="-e USER=$(id -un) -e LOGNAME=$(id -un) -e HOME=/tmp"
```

`scripts/emit_default_spec.py` builds its own `docker run` and sets the same three variables directly, since a caller cannot reach it from outside the skill.

`references/pipeline-and-state.md` mandated `--user` without naming what it requires alongside. It now points at the identity block and states that dropping `--user` is not the alternative — that reintroduces the root-owned output `--user` exists to prevent.

## Why only these three variables

Verified rather than assumed. `getpass.getuser()` reads `LOGNAME`, `USER`, `LNAME` and `USERNAME` before it consults `pwd`, so any non-empty name satisfies it — the name does not have to exist in the image. Once `HOME` is writable, `MPLCONFIGDIR` and `TORCHINDUCTOR_CACHE_DIR` resolve under it on their own, so they are redundant:

| | bare `--user` | `+ USER/LOGNAME/HOME` |
|---|---|---|
| inductor cache dir | `KeyError: getpwuid()` | `/tmp/torchinductor_<user>` |
| matplotlib config | `Permission denied: /.config` | `/tmp/.config/matplotlib` |
| cupy cache | `Permission denied: /.cupy` | ok |

Reproduced and re-verified on both images this skill uses — `tao-toolkit:7.1.0-pyt` and the pinned data-services nightly — by running as uid 159080. This is the shape the sibling DEFT skills (`tao-run-deft-aoi`, `tao-run-deft-aoi-cosmos3`, `tao-run-deft-iaa`) already ship.

## Verification

- exact traceback reproduced under `--user 159080:1000`, and cleared by the block
- all 21 `docker run` sites audited; the 11 that pass `--user` now carry `$DOCKER_IDENTITY`, the rest run as root and are unaffected
- `emit_default_spec.py` run end-to-end against the data-services image
- every fenced bash block in the skill parses; 827/827 state-machine assertions pass; `stamp_versions.py --check --strict-strays` clean

## Out of scope

Bug 6603226 also notes two other skills with the same defect. Both confirmed real, both left alone here since these bugs are filed against DEFT-OD and each deserves its own change:

- `tao-run-deft-aoi/references/prepare-for-inference.md:99`
- `tao-finetune-cosmos-reason/scripts/prepare_cosmos3_vlm_checkpoint.py:112`

Bug 6603226 also asks for a canonical rule plus a CI grep failing any `--user` launch with no identity, which is a bank-wide change and not attempted here.

### DIFF
--- a/skills/applications/tao-run-deft-object-detection/references/preflight.md
+++ b/skills/applications/tao-run-deft-object-detection/references/preflight.md
@@ -218,6 +218,38 @@ Resolve everything you can before asking the user. Parameter precedence is stric
 - `kpi.conf_threshold` — `0.0` (frozen as `config.kpi_conf_threshold`; every phase is scored at it)
 - workspace root — user prompt, else `~/workspace`
 
+## Container identity
+
+Every launch in this skill passes `--user "$(id -u):$(id -g)" $DOCKER_IDENTITY`. A uid that has no
+`/etc/passwd` entry inside the image — which is every uid but 1000 on the TAO images —
+then has no name, and `getpass.getuser()` falls back to `pwd.getpwuid()` and raises.
+torch calls it while setting up the inductor cache directory, so the container dies at
+import with `KeyError: 'getpwuid(): uid not found: <uid>'` before the requested action
+starts. `$HOME` is unset for the same reason and defaults to `/`, which is not writable,
+so matplotlib and cupy fail on their cache directories.
+
+Export the identity beside the mounts and pass it to every launch:
+
+```bash
+DEFT_HOME="$WORKSPACE/.deft_home"
+mkdir -p "$DEFT_HOME"
+DOCKER_IDENTITY="-e USER=deft -e HOME=$DEFT_HOME"
+export DOCKER_IDENTITY
+```
+
+Both are required, and neither is sufficient alone. `USER` satisfies
+`getpass.getuser()`, which reads it before consulting `pwd`; any non-empty value works,
+and a literal keeps container logs identical whoever runs the workflow. `HOME` defaults
+to `/`, which is not writable, so without it matplotlib and cupy fail on their cache
+directories even once the uid has a name.
+
+`HOME` points into the workspace rather than at `/tmp` so the caches are owned by the
+caller and removed with the workspace.
+
+This is invisible on a uid-1000 workstation and fires on every shared or cloud host, so a
+run that works locally proves nothing about it. Dropping `--user` is not the alternative:
+see `references/pipeline-and-state.md`.
+
 ## Container mounts
 
 Export them once, right after init, and use the variable in every launch:
--- a/skills/applications/tao-run-deft-object-detection/scripts/emit_default_spec.py
+++ b/skills/applications/tao-run-deft-object-detection/scripts/emit_default_spec.py
@@ -66,9 +66,14 @@ def emit_from_ds(module: str, image: str, workdir: Path) -> str:
     # one", and a leftover would otherwise be returned as a fresh success.
     spec.unlink(missing_ok=True)
 
+    # --user with a uid the image has no /etc/passwd entry for leaves the process
+    # nameless, and getpass.getuser() then raises before `default_specs` runs. USER is
+    # read before pwd is consulted, so any non-empty value settles it. HOME would
+    # default to "/", which is not writable; the workdir is already mounted.
     cmd = [
         "docker", "run", "--rm", "--gpus", "all", "--ipc=host",
         "--user", f"{os.getuid()}:{os.getgid()}",
+        "-e", "USER=deft", "-e", f"HOME={workdir}",
         "-v", f"{workdir}:{workdir}", "-w", str(workdir),
         image, module, "default_specs", f"results_dir={workdir}",
     ]
--- a/skills/applications/tao-run-deft-object-detection/references/grounding-dino.md
+++ b/skills/applications/tao-run-deft-object-detection/references/grounding-dino.md
@@ -172,7 +172,7 @@ pool's `classes.yaml`, not a class list. Pass it in every phase.
 No training at baseline. Score the user-supplied zero-shot / pretrained checkpoint:
 
 ```bash
-docker run --rm --gpus all --ipc=host --user "$(id -u):$(id -g)" \
+docker run --rm --gpus all --ipc=host --user "$(id -u):$(id -g)" $DOCKER_IDENTITY \
   -v "$WORKSPACE:$WORKSPACE" $EXTRA_MOUNTS -w "$WORKSPACE" \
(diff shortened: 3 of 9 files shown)


skills/applications/tao-run-deft-object-detection/references/preflight.md
```diff
@@ -218,6 +218,51 @@ Resolve everything you can before asking the user. Parameter precedence is stric
...
+
+```bash
+DEFT_HOME="$WORKSPACE/.deft_home"
+mkdir -p "$DEFT_HOME"
+DOCKER_IDENTITY="-e USER=deft -e LOGNAME=deft -e HOME=$DEFT_HOME"
+DOCKER_IDENTITY="$DOCKER_IDENTITY -e TORCHINDUCTOR_CACHE_DIR=$DEFT_HOME/.cache/inductor"
+export DOCKER_IDENTITY
+```
+
+`getpass.getuser()` reads `LOGNAME`, `USER`, `LNAME` and `USERNAME` before it consults
```
sfarhat-nvidia: So do all 4 of them need to be set? Right now 2 of them are so unclear

skills/applications/tao-run-deft-object-detection/references/preflight.md
```diff
@@ -218,6 +218,51 @@ Resolve everything you can before asking the user. Parameter precedence is stric
...
+`pwd`, so any non-empty name satisfies it — the name does not have to exist in the image.
+A fixed `deft` is used rather than the calling user's name: nothing derives a path from it
+once `HOME` and the cache directories are set explicitly, so the value only ever reaches
+container logs, and a literal keeps those identical whoever runs the workflow and keeps
+the operator's identity out of artifacts that get attached to bug reports.
+
+`HOME` points into the workspace rather than at `/tmp`. The workspace is already
```
sfarhat-nvidia: Is this blurb about /tmp necessary? Does it currently go to /tmp at the moment that we're trying to steer it away from this behavior? I thought it defaulted to /?

skills/applications/tao-run-deft-object-detection/references/preflight.md
```diff
@@ -218,6 +218,51 @@ Resolve everything you can before asking the user. Parameter precedence is stric
...
+bind-mounted and owned by the calling user, so the frameworks get somewhere writable
+either way — but the container's own `/tmp` dies with `--rm`, and every stage would then
+rebuild the same matplotlib, triton and inductor caches from scratch. On the workspace
+they persist across the run's dozen-odd containers, they survive for inspection when a
+stage misbehaves, and they are removed with the workspace rather than lingering in a
+shared `/tmp`. This follows `skills/platform/tao-run-on-docker`, which puts `HOME` on the
+writable mount for the same reasons.
+
+`TORCHINDUCTOR_CACHE_DIR` is set separately because torch's inductor cache does not follow
```
sfarhat-nvidia: Do we need this paragraph? Seems more like a justification for something you're doing earlier

skills/applications/tao-run-deft-object-detection/scripts/emit_default_spec.py
```diff
@@ -66,9 +66,18 @@ def emit_from_ds(module: str, image: str, workdir: Path) -> str:
...
+    # the container raises KeyError before `default_specs` runs. getuser() reads
+    # these env vars before it consults pwd, and any non-empty name satisfies it, so a
+    # literal is used rather than the caller's name -- nothing derives a path from it.
+    # HOME would otherwise default to "/", which is not writable. It points at the
+    # workdir, which is already mounted and writable, so the caches the frameworks
+    # build do not go to a container /tmp that --rm discards.
     cmd = [
         "docker", "run", "--rm", "--gpus", "all", "--ipc=host",
         "--user", f"{os.getuid()}:{os.getgid()}",
+        "-e", "USER=deft", "-e", "LOGNAME=deft", "-e", f"HOME={workdir}",
```
sfarhat-nvidia: Same thing here about if we need all 4 envvars set or not

arihantj-nv:
Thanks — all four comments were right, and the answer to the main one is that only two of the four were doing anything. Tested each combination against a uid absent from the image:

| set | result |
|---|---|
| nothing | \`KeyError: getpwuid(): uid not found\` |
| \`USER\` only | \`PermissionError: '/.cupy'\` |
| \`HOME\` only | \`KeyError: getpwuid()\` |
| **\`USER\` + \`HOME\`** | **works** |
| \`USER\` + \`LOGNAME\` + \`HOME\` | identical to the above |

So `USER` and `HOME` are both required and neither is sufficient alone; `LOGNAME` is redundant, since `getpass.getuser()` is satisfied by whichever of the four names it finds first. `TORCHINDUCTOR_CACHE_DIR` was not required either — without it the inductor cache goes to `/tmp`, which is writable and works fine; it was a cache-persistence optimisation, not part of the fix. Both are now dropped:

```bash
DOCKER_IDENTITY="-e USER=deft -e HOME=$DEFT_HOME"
```

On `/tmp` — you are right that it defaults to `/`, not `/tmp`. `/tmp` was an earlier version of this PR, so that blurb was justifying a road not taken. Removed. The prose is now one paragraph saying what each of the two variables is for, plus one line on why `HOME` sits in the workspace rather than `/tmp` (ownership and cleanup). The `TORCHINDUCTOR_CACHE_DIR` justification paragraph went with the variable.

`emit_default_spec.py` builds its own launch and got the same reduction.

sfarhat-nvidia (approved):
newest commit addressed all my concerns, lgtm

PR #151, opened 2026-08-19
[TAO-2230][Bugfix] deft-od: mount the detection files, and admit arbitrary paths

arihantj-nv:
Fixes NVBug 6635689.

## The defect

`tmm unique_neighbor_matching` opens both COCO detection files from inside the container under `class_stratified` allocation. `init_deft_state.py` recorded them in config and left them out of the derived mount list, so the stage failed on a path its own spec had named:

```
FileNotFoundError: [Errno 2] No such file or directory: ".../hl_coco_its_kpi.json"
  File ".../nvidia_tao_ds/core/utils/dataset_loading.py", line 144, in _load_coco
```

That file is the `target_detection_file` — the KPI detections, which live with the read-only dataset rather than in the per-run workspace. So this fires on the normal layout, not an unusual one.

The reporter also cites four overlays that mount only the workspace. Those were fixed in #124, which added `$EXTRA_MOUNTS` to every documented launch — but #124 did not add these two paths to the list that variable is built from, so the stage still failed.

## Audit

Every path-valued config key, against the derived mount list:

| | mounted |
|---|---|
| `zero_shot_checkpoint`, `train_spec_template` | yes |
| `source_pool_embeddings`, `source_pool_annotations`, `pool_images` | yes |
| `codetr_checkpoint`, `codetr_classmap` | yes |
| `embedding_model_path` (as repo root when local) | yes |
| `kpi_images_dir`, `ground_truth_labels_dir`, `class_mapping` | yes |
| `source_detection_file`, `target_detection_file` | **no** |

These two were the only gap.

## `--extra-mount`

The derivation can only cover paths the run's own inputs name. A run referencing anything else — a spec that points at an auxiliary checkpoint, a second dataset — had no remedy but editing the skill. `--extra-mount PATH` is repeatable, validated at init like any other input, and mounted by parent directory when it is a file.

Existence-only validation, deliberately: a mount may be a file or a directory, and `check_artifact` would reject a directory as "not a file".

## Verification

833/833 state-machine assertions (up from 827). New G26 covers: a target detection file outside the workspace is mounted; `--extra-mount` admits a path no config key names; a detection file inside the workspace adds no mount; a non-existent `--extra-mount` is refused and the message names the flag.

### DIFF
--- a/skills/applications/tao-run-deft-object-detection/references/preflight.md
+++ b/skills/applications/tao-run-deft-object-detection/references/preflight.md
@@ -234,6 +234,12 @@ Echo it and read the result. A command substitution that fails leaves `EXTRA_MOU
 empty and every later launch still runs, mounting nothing extra — the same silent
 failure this block exists to prevent.
 
+`init_deft_state.py` derives this list from the run's own inputs — the checkpoints, the
+KPI images and labels, the class mapping, the encoder repo, the pool, and both COCO
+detection files. A path that none of those name is invisible to it: pass
+`--extra-mount PATH` (repeatable) at init and it joins the list. Files are mounted by
+their parent directory.
+
 Every `docker run` in this skill is written as
 `-v "$WORKSPACE:$WORKSPACE" $EXTRA_MOUNTS -w "$WORKSPACE"`. With inputs inside the
 workspace it expands to nothing and the command is unchanged; with KPI data or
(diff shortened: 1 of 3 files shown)


arihantj-nv:
Superseded: #151 was the base of #152, so squash-merging #152 (37295c29) brought these commits into `main` with it.

Verified present on main: the detection files in the derived mount list, `--extra-mount`, and the completed pre-flight init block. This branch now diffs net-negative against main — merging it would revert #152's work — so closing rather than rebasing.

NVBug 6635689 is fixed by 37295c29.

PR #152, opened 2026-08-19
[TAO-2230][Bugfix] deft-od: Overlay and Set Default Fixes

arihantj-nv:
Fixes NVBug 6635690, the third of its three items. The other two are already correct on `main`:

- the KPI command is `analytics kpi_analyze` in both the workflow and the leaf skill
- the embed `model` field is `SigLIP`, a type name; `model_path` is the separate field that takes the HuggingFace id

## The defect

`init_deft_state.py` froze `kpi_conf_threshold` into config and the documented spec build never passed it, while `assets/overlays/kpi_analyze.yaml` pinned `kpi.conf_threshold` outright. The recorded value therefore had no effect on any metric:

```
deft_state.json     "kpi_conf_threshold": 0.3
kpi_overrides.json  "kpi.conf_threshold": {"source": "kpi_analyze.yaml", "value": 0.0}
```

Nothing errored, because nothing ever tried to set it. `apply_spec_overrides.py` rejects a `--set` that collides with a workflow default and names `--allow-workflow-default-override` — that guard worked correctly; the wiring was simply absent. So `deft_state.json` advertised a parameter the run did not honour, and two differently-configured runs recorded identical settings.

## The fix

`kpi.conf_threshold` is a run parameter, not a workflow constant — init has a `--kpi-conf-threshold` flag for it. It moves out of the overlay and into `--set`, read from `config.kpi_conf_threshold`:

```bash
--set kpi.conf_threshold=<state.config.kpi_conf_threshold> \
```

Leaving it in the overlay and passing `--allow-workflow-default-override` instead was the alternative, and it is worse: that flag would fire on every run and stop meaning anything. Things the run owns belong in `--set`; things the workflow requires belong in the overlay. This one was in the wrong bucket.

## The default is 0.0, not 0.3

`0.0` is the value the loop should score at. Inference writes its labels at `0.0`, so scoring there covers the full PR curve; a higher threshold truncates the low-confidence tail and yields a number no run scored at `0.0` can be compared against. The default therefore moves from `0.3` to `0.0`, which is also what the loop has always actually done — so **no reported metric changes**, and existing results stay comparable.

`--kpi-conf-threshold` is how a run that needs a different operating point says so once, at init, and has every phase honour it.

The reporter now states the threshold whenever it is not `0.0`, so a reader can see when a run is not comparable.

## Verification

Traced end-to-end against the real data-services image — user supplies `0.3` at init, it lands in `deft_state.json`, and the documented spec build puts it in the spec with no override flag needed:

```
init rc: 0
deft_state.json config.kpi_conf_threshold = 0.3
apply rc: 0 | needed --allow-workflow-default-override? NO
spec kpi.conf_threshold = 0.3
overlay values intact: num_recall_points 11, ignore_sqwidth 40, iou_threshold 0.5
```

830/830 state-machine assertions (up from 827). New G26 covers the `0.0` default, the recorded override, and the `[0, 1]` bound.

### DIFF
--- a/skills/applications/tao-run-deft-object-detection/agents/reporter.md
+++ b/skills/applications/tao-run-deft-object-detection/agents/reporter.md
@@ -103,7 +103,11 @@ Include the `tokens` column only when the field is present.
 ## 4. Configuration
 
 Encoder, allocation policy, rare classes, mining multiplier, per-class AP50
-thresholds, epochs, learning rate, GPU count.
+thresholds, epochs, learning rate, GPU count, and `kpi_conf_threshold`.
+
+State the KPI confidence threshold whenever it is not `0.0`. Every mAP in the
+report is scored at it, and a run scored anywhere else is not comparable to one
+scored at `0.0` — a reader who cannot see the value cannot know that.
 
 ## 5. Observations
 
--- a/skills/applications/tao-run-deft-object-detection/assets/overlays/kpi_analyze.yaml
+++ b/skills/applications/tao-run-deft-object-detection/assets/overlays/kpi_analyze.yaml
@@ -2,8 +2,12 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 # `analytics kpi_analyze` — the settings that do not vary per run, applied over
-# the spec `analytics default_specs` emits. Paths, `visualize.tag` and
-# `data.kpi_sources` are run-specific and stay as --set.
+# the spec `analytics default_specs` emits. Paths, `visualize.tag`,
+# `data.kpi_sources` and `kpi.conf_threshold` are run-specific and stay as --set.
+#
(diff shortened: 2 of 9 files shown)


arihantj-nv:
Rebased onto `dev/od-mount-detection-files` (#151) so the two no longer conflict.

Both PRs appended a gate at the same anchor in `scripts/tests/test_state_machine.sh`, so whichever merged second would have hit a conflict. #152 now contains #151's commits, and the gates are in order: G26 (mounts) then G27 (KPI threshold). 836/836 assertions.

Either merge order is now safe. Merging #151 first leaves #152 as its own single commit; merging #152 first brings #151 along with it.

ddevendran-nvidia (approved):

PR #153, opened 2026-08-19
[TAO-2230][Bugfix] deft-od: name the modules an interpreter is missing

arihantj-nv:
## Why

On a host missing one dependency, `deft_python.sh` reported the whole list of six. The reader is sent looking for five modules that are already installed, and the skill's own instruction is to report what is missing and stop.

## What

- Report, per candidate interpreter, only the modules that interpreter actually lacks.
- Build the probe and the report from one list so the two cannot drift.
- Selection is unchanged; the per-module check runs only on the failure path.


sfarhat-nvidia (approved):
lgtm

PR #65, opened 2026-07-29
ci(smoke): tao-skill-eval-OSS /build trigger test (DO NOT MERGE)

christinayyw:
Smoke test for the Colossus Jenkins agent + Blossom `/build` -> `tao-skill-eval-OSS` path. Checkout-only. **Do not merge.**

PR #154, opened 2026-08-19
Re-sign all 62 skills on release/7.1.0

christinayyw:
## What
Force re-sign **all 62 skills** on `release/7.1.0` (base + merge target = `release/7.1.0`, not main).

## Why
Post GitLab→GitHub migration, signatures went stale, so the NVSkills catalog holds skills on their previous signed copy. Re-signing every skill against current content refreshes them. main has diverged too far to sign there, so we sign directly on the release branch (which is the intended catalog sync source).

## How
Trailing-newline nudge on each of the 62 SKILL.md files brings them into the changed-set; `/nvskills-ci` re-signs them. Dispatch token was refreshed (prior 403 was a 2FA-compliance block on the old token account).

Will fan out `/nvskills-ci` re-triggers if the backend caps skills-per-run.


christinayyw:
/nvskills-ci

Re-signing all 62 skills — #65 merged (release/latest now allowlisted), token fixed.

christinayyw:
/nvskills-ci

Track 1: re-scoped to the 34 gate-clean skills (0 Tier-1 high-risk) so gate:content passes. The 28 with high-risk findings will follow in a separate remediation PR.

PR #160, opened 2026-08-20
[TAO-0000][Feature] video_clip: repoint to official pyt image, retire interim dev container

christinayyw:
## What

The 7.2.0 pyt RC — `nvcr.io/nvstaging/tao/tao-toolkit-pyt:7.2.0-rc-45-multiarch` — is a full base-image rebuild carrying [tao-pytorch !620](https://gitlab-master.nvidia.com/nvidia-tao-toolkit/tao-pytorch/-/merge_requests/620), so it now ships the complete `video_clip` stack (`model.backbones` packaged **and** `decord==0.6.0`). This retires the interim IV2CLIP dev container (`v7.0.1-pyt2.1.0-py3-03`) and cleans up every mention of it.

## Changes

- **`versions.yaml`** — remove the `images.tao_toolkit.video_clip` key (the interim nvstaging pin). The skill now stamps from `images.tao_toolkit.pyt`.
- **`references/skill_info.yaml`** + **`SKILL.md`** — repoint `container_image` / `VIDEO_CLIP_IMAGE_DEFAULT` to the pyt tag via the `# versions-key: images.tao_toolkit.pyt` marker.
- **`SKILL.md`** — drop the interim release note, the known-broken-image note (`py3-01/02/03` missing `model.backbones`), and the `decord` derived-image workaround. A generic data-free import-smoke preflight is retained as a health check.

## Validation

- `scripts/stamp_versions.py --check` passes (all versions-key stamps match `versions.yaml`, 0 stray warnings).
- The `decord` / `model.backbones` workarounds were removed on the premise that `7.2.0-rc-45-multiarch` is a full base rebuild incorporating !620. **skill-eval CI on this PR validates the image end-to-end** — if a runtime dep is missing, the eval will surface it as a container/image defect (route to the pyt image owner), which is the intended artifact-vs-skill behavior.

Signed-off-by: Christina Wang <christwang@nvidia.com>

PR #161, opened 2026-08-20
stamp: scan scripts/ so duplicated pins can't drift unnoticed

christinayyw:
## Why

The cosmos-rl unit tests broke on `release/7.2.0` (blocking unrelated PRs, e.g. #160). Root cause: the build-13 versions.yaml bump moved `images.tao_toolkit.cosmos_rl`; the stamper fanned it into the **marked** skill + backend sites, but `scripts/tests/test_resolve_tao_image.py` duplicated the pin as an **unmarked literal** the stamper never scanned — its scope was `skills/` + `templates/` only. So it drifted silently and the marked-drift/stray check couldn't see it.

(The immediate red was already cleared by realigning the constant directly on `release/7.2.0`. This PR fixes the **stamping logic** so it can't recur.)

## What

- **Scan `scripts/`** in addition to `skills/` and `templates/`, so marked pins and unmarked strays in CI helpers and test fixtures are covered. The versions-key tooling itself (`stamp_versions.py`, `migrate-to-version-keys.py`) is skipped — its docstrings document the marker format with example pins.
- **Fixes a genuine latent drift** the new scope surfaced: `scripts/ci/render_and_validate_templates.py` carried a marked `pyt` pin stuck at `7.1.0-pyt`; re-stamped to the current pin.
- **Marks the cosmos-rl test constant** with `# versions-key: images.tao_toolkit.cosmos_rl`, so it's now auto-managed like every other cosmos-rl site — a future bump re-stamps it, so this class of drift can't recur.
- **Annotates deliberate test fixtures** in `test_tao_job_record.py` with `# unpinned: test fixture`.

## Verification

- `python3 scripts/stamp_versions.py --check --strict-strays` → `OK … (0 stray warning(s))` over `skills/` + `templates/` + `scripts/`.
- All touched files compile; the 3 cosmos-rl tests' invariant holds (resolve == constant == skill == versions.yaml, all `cosmos_rl:7.2.0-rc-241-multiarch`).
- Net effect: a new unmarked image/wheel pin anywhere in those trees now fails `--check --strict-strays`.

> Follow-up: this hardening should also land on `main` so the next release cut inherits it (this PR targets `release/7.2.0` to fix the branch that's currently red).

Signed-off-by: Christina Wang <christwang@nvidia.com>

### DIFF
--- a/scripts/ci/render_and_validate_templates.py
+++ b/scripts/ci/render_and_validate_templates.py
@@ -46,7 +46,7 @@ COMMON = {
     "NUM_NODES": "2",
     "GPUS_PER_NODE": "8",
     "NUM_GPUS": "8",          # single-node templates spell it this way
-    "IMAGE": "nvcr.io/nvidia/tao/tao-toolkit:7.1.0-pyt",  # versions-key: images.tao_toolkit.pyt
+    "IMAGE": "nvcr.io/nvstaging/tao/tao-toolkit-pyt:7.2.0-rc-45-multiarch",  # versions-key: images.tao_toolkit.pyt
     "COMMAND": "dino train -e /data/specs/spec.yaml",
 }
 K8S_VALS = {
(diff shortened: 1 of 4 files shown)


christinayyw:
Superseded by #162, which targets `main` (canonical home for the stamper) with the `release/7.2.0` cherry-pick label. The cosmos-rl constant was already realigned on `release/7.2.0` directly; render_and_validate re-stamp on release will be handled at cherry-pick time (run `scripts/stamp_versions.py`).

PR #167, opened 2026-08-20
[TAO-2230][Bugfix] deft-od: make the documented commands runnable as written

arihantj-nv:
## Why

Several documented commands could not be run as written, and each failed in a way that pointed away from the documentation.

- Every stage names its spec in a shell variable that nothing assigned. An unset variable expands to empty, so `-e ""` reaches TAO as a blank spec and reads as a spec bug.
- Three values had two names each, so defining one left the other empty. One was live: pre-flight resolved the encoder into a variable init never read.
- The documented waits paired an artifact with a status file. The wait ends on the first condition that holds and the artifact appears at startup, so a stage was treated as finished when it had barely begun.
- `gap_analysis` had no documented spec build, and prep's embed step deferred to a spec written for a different stage.

## What

- Bind every spec variable beside the path its own reference documents, and settle on one name per value.
- Add a pre-flight block naming the run variables, so the set is stated once.
- Wait on `status.json` alone for stages whose outputs appear before they finish; `await_stage.py` now says so too.
- Document the `gap_analysis` and prep-embed spec builds.

Verified by running the documented commands against the real images.

### DIFF
--- a/skills/applications/tao-run-deft-object-detection/references/grounding-dino.md
+++ b/skills/applications/tao-run-deft-object-detection/references/grounding-dino.md
@@ -117,8 +117,10 @@ nobody applies. They are therefore held in `assets/overlays/grounding_dino_infer
 and applied on every run:
 
 ```bash
+INFER_SPEC="${RESULTS_DIR}/<phase>/infer_grounding_dino.yaml"
+
 <skill_root>/scripts/deft_python.sh <skill_root>/scripts/apply_spec_overrides.py \
-  --spec "${RESULTS_DIR}/<phase>/infer_grounding_dino.yaml" \
+  --spec "$INFER_SPEC" \
   --apply-workflow-defaults <skill_root>/assets/overlays/grounding_dino_inference.yaml \
   --set inference.checkpoint=<checkpoint> \
   --set inference.num_gpus="$NUM_GPUS" \
@@ -153,8 +155,8 @@ Nothing about this is enforced by TAO, so check it before every inference:
 
 ```bash
 <skill_root>/scripts/deft_python.sh <skill_root>/scripts/verify_class_contract.py \
-  --inference-spec "$GDINO_INFER_SPEC" \
-  --kpi-mapping "$KPI_CLASS_MAPPING" \
+  --inference-spec "$INFER_SPEC" \
+  --kpi-mapping "$CLASS_MAPPING" \
   --state "${RESULTS_DIR}/deft_state.json" \
   --labelmap "${RESULTS_DIR}/iter${N}/tmm/annotations/labelmap.json"
 ```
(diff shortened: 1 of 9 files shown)


skills/applications/tao-run-deft-object-detection/scripts/await_stage.py
ddevendran-nvidia, line 153: According to Codex this might be a bug during a retry. If the workflow is retried then there is a possible "success" from an older run, and this status_matches will take that "success" as an indication that the stage finished. This only matters if it retries from an earlier stage

ddevendran-nvidia: This is what Codex said: The reason I flagged it is that the docs explicitly say --newer-than prevents a previous attempt’s status from satisfying the wait. As implemented, it only partially does that: it guards the file mtime, not the matching line. So the sharper review comment might be: either document that successful stage dirs must not be reused, or make await_stage.py ignore old status lines.

skills/applications/tao-run-deft-object-detection/references/tao-analyze-detection-kpi.md
```diff
@@ -30,12 +30,15 @@ Do not hand-write it. Emit `analytics default_specs`, then apply the overlay —
...
 <skill_root>/scripts/deft_python.sh <skill_root>/scripts/emit_default_spec.py \
   --stage kpi_analyze --ds-image "$TAO_DS_IMAGE" \
-  --out "${RESULTS_DIR}/<phase>/kpi_spec.yaml"
+  --out "$KPI_SPEC"
 
 <skill_root>/scripts/deft_python.sh <skill_root>/scripts/apply_spec_overrides.py \
-  --spec "${RESULTS_DIR}/<phase>/kpi_spec.yaml" \
+  --spec "$KPI_SPEC" \
   --apply-workflow-defaults <skill_root>/assets/overlays/kpi_analyze.yaml \
   --set results_dir="${RESULTS_DIR}/<phase>/kpi" \
```
ddevendran-nvidia, line 43: Shouldn't this be ${PHASE}?

skills/applications/tao-run-deft-object-detection/references/grounding-dino.md
```diff
@@ -117,8 +117,10 @@ nobody applies. They are therefore held in `assets/overlays/grounding_dino_infer
 and applied on every run:
 
 ```bash
+INFER_SPEC="${RESULTS_DIR}/<phase>/infer_grounding_dino.yaml"
```
ddevendran-nvidia, line 120: Shouldn't this be ${PHASE}?

ddevendran-nvidia (approved):
Had Codex Sol 5.6 review and it only found two potential issues:

* There are a bunch of <phase> that should be replaced with the PHASE variable, right? Check that all the variables are used where they should be.
* Also see Codex's comment about await_stage.py

Neither of these are blockers b/c they are more about durability than functionality. But check these.

PR #168, opened 2026-08-20
[TAO-2230][Bugfix] deft-od: record the KPI mAP instead of capturing it from stdout

arihantj-nv:
## Why

`kpi_analyze` prints the aggregate mAP to stdout and writes it nowhere, and `kpi_calc.csv` has one row per class with no aggregate row.  Now we save as an artifact which is used throughout. 

## What

- `summarize_kpi.py` recomputes the aggregate from the csv — it is the unweighted mean of the per-class APs — and writes `kpi_summary.json` beside it. No dependency on stdout.
- Use the `class_name` column where the image has it (tao-data-services#31), labelling each AP and dropping any `Summary` row. Fall back to the log where it does not; the pinned image predates that change.
- The reporter reads the summary first.

Verified against a full 5k run: the derived values match what the stage printed.


ddevendran-nvidia (approved):
A very minor possible issue from Codex (but only P2):

`The documented MAP_VALUE=$(... | tail -1) hides summarize_kpi.py failures. Without set -o pipefail, the pipeline exits with tail’s status, so a missing/bad CSV gives exit 0 and MAP_VALUE='': [tao-analyze-detection-kpi.md#L169-L172](https://github.com/NVIDIA-TAO/tao-skill-bank/blob/a2ae349c95aade43d1c10d8cb8a1e0122193bc05/skills/applications/tao-run-deft-object-detection/references/tao-analyze-detection-kpi.md#L169-L172). I reproduced that locally. The later commit_stage.py --map-value "$MAP_VALUE" should fail, but the failure points at commit rather than at the KPI summary step. Safer: capture script output first, check status, then take the last line, or require set -o pipefail.`

Approving b/c it's only P2

PR #169, opened 2026-08-20
[TAO-2230][Bugfix] deft-od: name the containers each stage launches

arihantj-nv:
## Why

The documented launches passed no `--name`, so their containers took Docker's generated names. On a shared host the run cannot tell its own containers from anyone else's, and cleanup after a failure is guesswork.

## What

Name each stage's container after that stage, matching the form the KPI stage already used.

### DIFF
--- a/skills/applications/tao-run-deft-object-detection/references/grounding-dino.md
+++ b/skills/applications/tao-run-deft-object-detection/references/grounding-dino.md
@@ -174,7 +174,7 @@ pool's `classes.yaml`, not a class list. Pass it in every phase.
 No training at baseline. Score the user-supplied zero-shot / pretrained checkpoint:
 
 ```bash
-docker run --rm --gpus all --ipc=host --user "$(id -u):$(id -g)" $DOCKER_IDENTITY \
+docker run --rm --name "deft_${PHASE}_infer" --gpus all --ipc=host --user "$(id -u):$(id -g)" $DOCKER_IDENTITY \
   -v "$WORKSPACE:$WORKSPACE" $EXTRA_MOUNTS -w "$WORKSPACE" \
   "$TAO_PYT_IMAGE" \
   grounding_dino inference -e "$INFER_SPEC" \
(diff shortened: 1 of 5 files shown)


skills/applications/tao-run-deft-object-detection/references/grounding-dino.md
```diff
@@ -174,7 +174,7 @@ pool's `classes.yaml`, not a class list. Pass it in every phase.
 No training at baseline. Score the user-supplied zero-shot / pretrained checkpoint:
 
 ```bash
-docker run --rm --gpus all --ipc=host --user "$(id -u):$(id -g)" \
+docker run --rm --name "deft_${PHASE}_infer" --gpus all --ipc=host --user "$(id -u):$(id -g)" \
```
ddevendran-nvidia: Is this dependent on seeing the other reference? Because ${PHASE} doesn't seem to be in this reference. Also is N defined in this reference?

arihantj-nv:
Good catch, and it is worse than the line you flagged — thank you.

`${PHASE}` was defined only in `tao-analyze-detection-kpi.md`, and `${N}` was defined **nowhere in the skill**. `N` is used 90 times across 8 references. And since `SKILL.md` says to *"read only the current stage's overlay, then invoke"*, a definition in a sibling reference would not have reached a reader of `grounding-dino.md` anyway.

So this was not a naming-PR problem so much as a pre-existing hole that the container names made visible. Fixed one PR down, in #167, which is the one about undefined variables:

- Pre-flight now defines both alongside the run-wide variables — `N` is the iteration counter, `PHASE` is `baseline` or `iter${N}`, re-exported at the top of each iteration.
- `SKILL.md` states them next to the read-one-overlay rule, since that file is always read even when only one overlay is.
- The init block's own inputs are listed too, which had the same gap.

Worth recording why my earlier sweep for undefined variables missed this: the pattern required three characters or more, so a single-letter `N` was never checked. Re-run with single letters included, and the remaining hits are all false positives (`CODETR` is assigned via `&&` on one line, `HF_HOME` is a passthrough).

#168 and #169 have been restacked on top.

ddevendran-nvidia (approved):

PR #171, opened 2026-08-21
[TAO-2546][Bugfix] DEFT CR ITS Workflow Skill: Cosmos RL Runtime

ddevendran-nvidia:
This PR consolidates the bugs around the Cosmos RL issues:
* It no longer uses the Cosmos Reason skill because the cosmos framework changes confused the agent. Instead, it states the commands that the agent should run. As a result, the broken runtime hook resolution in **bug 6630387** is no longer there (due to the CR skill).
* It sets the `TAO_API_RESULTS_DIR`, `TAO_STATUS_FILE`, and `TAO_API_JOB_ID` to fix **bug 6630386**
* It automatically restores host permissions after train and evaluate (since TAO CR uses root permissions to write checkpoints). This fixes **bug 6630388**
* It fixes a new bug: the checkpoint symlink was never removed in the checkpoint cleanup
* If it detects an omni checkpoint it asks the user to use the CR skill to convert it into a reasoner checkpoint. This separates the two actions, and patches **bug 6630303**
* It calculates the number of training steps before each train and exits if the number of training steps is zero. It explicitly tells the user why. This fixes **bug 6630304**

Yes, this is for the TAO 7.2 DEFT CR3 ITS workflow

Had the agent run on the text only workflow but for a single iteration. None of the patches (like the TAO_API one) showed up. Asked it to check the location of the status.json to verify it dropped where we expected.

For the training steps one, asked the agent to run on a smaller subset of 26 samples. It recreated the original bug (it originally expected 3 steps though so its calculations were incorrect). Then gave it the patch

### Release note

```release-note
Fixed bugs for the DEFT CR ITS workflow related to Cosmos Reason runtime.
```

### Checklist

- [x] All tests pass locally
- [x] I added or updated documentation (README, docstrings, docs pages, examples)
- [x] I updated the version, changelog, or migration notes if this change requires it
- [x] If this touches components that are optional to install, the imports are guarded
      so the package still works without them (reviewers: please verify this)
- [x] I understand this contribution is licensed under the repository's license, and that
      my commit author name and email become permanently public once merged


sfarhat-nvidia (approved):
walked through this in person with Dharshi, lgtm

PR #196, opened 2026-08-26
[TAO-0000][Bugfix] DEFT CR ITS QA Round 2 Fixes

ddevendran-nvidia:

This fixes the following reported bugs:
* [6636959](https://nvbugswb.nvidia.com/NVBugs5/redir.aspx?url=/6636959): specifies the Docker command for Cosmos Embed because the skill's Docker command added results_dir and that was outputting data to the wrong spot
* [6636954](https://nvbugswb.nvidia.com/NVBugs5/redir.aspx?url=/6636954): makes it explicit that 130 is not necessarily a failure. It has the agent verify the outputs are correct in this case and, if they are correct, proceed.
* [6630438](https://nvbugswb.nvidia.com/NVBugs5/redir.aspx?url=/6630438): added numpy to the dependency list and checked if pip is installed and added it to the necessary dependencies
* Also made it explicit that gap analysis requires GPUs (only for the Data Services container and not the actual action)

It fixes bugs in the DEFT CR ITS workflow

Asked the agent to "Fine-tune Cosmos Reason using DEFT for ITS using the workspace ~/deft_workspace and workflow_text_bugfix_embedgen_smoketest.yaml" This workflow yaml sets the number of epochs to 1, number of neighbors to 1, and only requires generating embeddings for the KPI dataset. The run is fast as a result.

### Release note

```release-note
Fixed bugs in DEFT CR ITS workflow
```

### Checklist

- [ ] My commits are signed off per the DCO (`git commit -s`) — see `CONTRIBUTING.md`
- [ ] I have read the contributing guidelines
- [ ] The code follows the project's style, and lint and format checks pass locally
- [ ] I added or updated tests covering this change
- [ ] All tests pass locally
- [ ] I added or updated documentation (README, docstrings, docs pages, examples)
- [ ] I updated the version, changelog, or migration notes if this change requires it
- [ ] If this touches components that are optional to install, the imports are guarded
      so the package still works without them (reviewers: please verify this)
- [ ] No secrets, credentials, proprietary data, or customer data are included in this PR
- [ ] I understand this contribution is licensed under the repository's license, and that
      my commit author name and email become permanently public once merged

### DIFF
--- a/skills/data/tao-analyze-gaps-vlm-bcq/references/skill_info.yaml
+++ b/skills/data/tao-analyze-gaps-vlm-bcq/references/skill_info.yaml
@@ -1,6 +1,10 @@
 network_arch: tao-analyze-gaps-vlm-bcq
 type: data
 container_image: nvcr.io/nvidia/tao/tao-toolkit:7.1.0-data-services  # versions-key: images.tao_toolkit.data_services
+# Fixed compute requirement: request one GPU from the platform. Even though this
+# function doesn't need a GPU, the Data Services image always calls nvidia-smi.
+# This remains null because the strict VLM BCQ spec has no GPU-count field;
+# gpu_spec_key is a spec-field pointer, not a GPU count or device ID.
 gpu_spec_key: null
 required_credentials: []
 actions:
--- a/skills/applications/tao-run-deft-cr-its-mining/SKILL.md
+++ b/skills/applications/tao-run-deft-cr-its-mining/SKILL.md
@@ -184,7 +184,7 @@ The KPI parquet contains only the selected modality or modalities. The train par
 
 ### Run Cosmos Embed Inference
 
-Use the `tao-finetune-cosmos-embed` skill's inference action for each generated inference spec. That model skill owns the Cosmos Embed container image, action command, credentials, and platform-specific mount behavior. This workflow skill owns only the generated specs and the downstream conversion into mining-ready parquet files. Do not run inference or conversion for a dataset whose combined Parquet was supplied and staged during preparation. See `references/mining-loop.md` for exact commands, permission repair, and conversion.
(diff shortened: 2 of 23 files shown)


skills/data/tao-analyze-gaps-vlm-bcq/references/skill_info.yaml
```diff
@@ -1,6 +1,10 @@
 network_arch: tao-analyze-gaps-vlm-bcq
 type: data
 container_image: nvcr.io/nvidia/tao/tao-toolkit:7.1.0-data-services  # versions-key: images.tao_toolkit.data_services
+# Fixed compute requirement: request one GPU from the platform. Even though this
+# function doesn't need a GPU, the Data Services image always calls nvidia-smi.
```
sfarhat-nvidia, line 5: This is interesting; I think removing the nvidia-smi call altogether is more complicated than we'd like, but does the image simply fail altogether if nvidia-smi fails? Could this be error-handled (either at the data-services level or here at the skill level)? Not a blocker, but from a user perspective, requiring a gpu for a step that will have 0% utilization is not ideal

sfarhat-nvidia (commented):
only one issue/question, otherwise lgtm

skills/data/tao-analyze-gaps-vlm-bcq/references/skill_info.yaml
ddevendran-nvidia: Yeah, the image fails if you call the Docker without any GPUs :-( QA had to file a bug for it: https://nvbugspro.nvidia.com/bug/6670065 . The right thing would be to fix Data Services so it doesn't do that, but that's a larger fix and out-of-scope for this release unfortunately.

sfarhat-nvidia (approved):
as long as the image gets updated once release, good with me. approved

PR #201, opened 2026-08-26
fix: harden TAO workflow validation and Brev execution

ramanathan831:
## Summary

- normalize LoRA target-module inputs and propagate Cosmos runtime identity, dimensions, GPU topology, and sealed-plan Docker launch details
- fail closed for undecodable or non-canonical media, checkpoint tensor-key drift, and incomplete launch hardware data
- harden Framework checkpoint and metric helpers without changing the Video-CLIP template owned by #203
- make Brev readiness tolerant of appended CLI output and make the normative submit contract idempotent

## Root-cause review updates

- Framework checkpoint verification now reads tensor keys from actual safetensors headers for both indexed/sharded and single-file checkpoints. Indexed metadata is also checked against the referenced shard headers. Missing, malformed, duplicate, escaping, or mismatched weights fail closed. This fixes the verification boundary itself; it does not rename files, trust metadata hashes alone, or special-case the reported tensor name.
- Docker GPU resolution now inventories physical index, UUID, and memory, applies `CUDA_VISIBLE_DEVICES` using the same index/UUID-prefix convention as shared launch preflight, filters capacity within that visible set, records the selected host IDs in the sealed plan, and uses the exact device request in preflight and final Docker launch. It does not infer availability from physical `nvidia-smi` count or expose all GPUs after approving a subset.
- Full media decode validation is preserved, but ffprobe work is dispatched through a bounded eight-worker pool. I intentionally avoided sampling because it could let corrupt unsampled clips reach a long-running job.
- The Video-CLIP export-template hunk and its test were removed; #203 remains the single owner of that complete deploy change.

## NVBugs

6671248, 6671239, 6670478, 6670470, 6670461, 6670213, 6670029, 6669930, 6669910, 6669898, 6669893, 6669862, 6669859, 6669829, and duplicate 6669848.

## Validation

- `python -m pytest -q scripts/tests/test_cosmos_skill_workflow.py scripts/tests/test_launch_preflight.py scripts/tests/test_platform_contract.py`: **202 passed**
- focused review cases covering single-file/indexed tensor drift, CUDA-visible index/UUID selection, ineligible visible GPUs, exact Docker device rendering, and media preflight: **14 passed**
- `./scripts/validate-skills.sh`: passed
- Python compilation and `git diff --check`: passed
- live Brev readiness gate exited 0 with the CLI's appended instance-name output
- two identical guarded live submits produced one container; the test container was removed afterward

### DIFF
--- a/skills/models/tao-finetune-cosmos-reason/scripts/cosmos_common.py
+++ b/skills/models/tao-finetune-cosmos-reason/scripts/cosmos_common.py
@@ -15,7 +15,9 @@ import hashlib
 import json
 import os
 import re
+import shutil
 import statistics
+import subprocess
 import sys
 import tempfile
 from pathlib import Path
@@ -27,6 +29,7 @@ MODEL_ID_RE = re.compile(r"^[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+$")
 ACCURACY_TASKS = {"bcq", "binary", "mcq", "binary_choice", "multiple_choice"}
 DATASET_FAMILIES = {"auto", "video_conversation", "task_aware_video_reasoning"}
 MEDIA_FIELDS = ("video", "video_id", "image", "image_id", "media", "media_path")
+TAO_VL_MEDIA_FIELDS = ("video_id", "image_id")
 CLASSIFICATION_LABEL_SETS = (
     frozenset({"a", "b", "c", "d"}),
     frozenset({"yes", "no"}),
@@ -222,6 +225,27 @@ def _record_media(record: Mapping[str, Any]) -> list[str]:
     return values
 
 
+def decode_media(path: Path) -> None:
+    """Decode-probe one frame so corrupt video inputs fail before launch."""
+    if path.suffix.casefold() not in {".mp4", ".mov", ".mkv", ".avi", ".webm", ".m4v"}:
+        return
+    ffprobe = shutil.which("ffprobe")
+    if not ffprobe:
+        raise WorkflowError(
+            "media decode validation requires ffprobe on the inspection host; "
+            f"cannot validate {path}"
+        )
+    completed = subprocess.run(
+        [ffprobe, "-v", "error", "-select_streams", "v:0", "-read_intervals", "%+#1",
+         "-show_entries", "frame=media_type", "-of", "csv=p=0", str(path)],
+        text=True, capture_output=True, check=False,
+    )
+    if completed.returncode or "video" not in completed.stdout.split():
+        detail = completed.stderr.strip().splitlines()
+        reason = detail[-1] if detail else "no decodable video frame"
+        raise WorkflowError(f"media decode validation failed for {path}: {reason}")
+
+
 def _record_key(record: Mapping[str, Any]) -> str:
     identity = {
         "id": record.get("id") or record.get("sample_id") or record.get("video_id"),
@@ -361,8 +385,17 @@ def inspect_dataset(
                     schema_errors.append(f"{annotation_path}:{index}: task-aware record has no task")
                 if requested_tasks and task not in requested_tasks:
                     continue
-                if not _record_media(record):
-                    schema_errors.append(f"{annotation_path}:{index}: task-aware record has no media field")
+                canonical_media = [
+                    field for field in TAO_VL_MEDIA_FIELDS
+                    if isinstance(record.get(field), str) and record.get(field)
+                ]
+                if len(canonical_media) != 1:
+                    aliases = [field for field in MEDIA_FIELDS if record.get(field)]
+                    schema_errors.append(
+                        f"{annotation_path}:{index}: task-aware tao-vl-reason-v1.0 "
+                        "record requires exactly one canonical video_id or image_id; "
+                        f"found {aliases or 'none'}"
+                    )
                 conversations = record.get("conversations") or record.get("messages")
                 has_conversation_target = isinstance(conversations, list) and len(conversations) >= 2
                 has_question_answer = (
--- a/skills/models/tao-finetune-cosmos-reason/scripts/cosmos_workflow.py
+++ b/skills/models/tao-finetune-cosmos-reason/scripts/cosmos_workflow.py
@@ -32,6 +32,7 @@ import yaml
 from cosmos_common import (
     WorkflowError,
     assert_no_overlap,
+    decode_media,
     dataset_parity,
     inspect_dataset,
     inspect_model,
@@ -1945,6 +1946,7 @@ def _env(
     val_media: Sequence[str],
     rl_video_runtime: Mapping[str, Any] | None = None,
     framework_video_runtime: Mapping[str, Any] | None = None,
+    model_profile: Mapping[str, Any] | None = None,
 ) -> dict[str, str]:
     tao_job_id = args.tao_job_id or args.experiment_id
     status_path = str(Path(args.container_results_dir) / tao_job_id / "status.json")
@@ -1990,6 +1992,10 @@ def _env(
                     "baked container path below /tao-patches-framework-*"
                 )
             common["PYTHONPATH"] = str(overlay)
+        resolved_profile = model_profile or {
+            "frame_width": args.video_resized_width or args.video_frame_width or 1280,
+            "frame_height": args.video_resized_height or args.video_frame_height or 720,
+        }
         common.update(
             {
                 "PYTHONNOUSERSITE": "1",
@@ -2010,6 +2016,8 @@ def _env(
                 "TAO_VIDEO_VAL_MEDIA": val_media[0],
                 "TAO_VIDEO_VAL_MEDIA_ROOTS": json.dumps(framework_val_media),
                 "TAO_VIDEO_NUM_FRAMES": str(args.frames),
+                "TAO_VIDEO_FRAME_WIDTH": str(resolved_profile["frame_width"]),
+                "TAO_VIDEO_FRAME_HEIGHT": str(resolved_profile["frame_height"]),
                 "TAO_VIDEO_SYSTEM_PROMPT": args.system_prompt,
                 "TAO_VIDEO_CACHE_SIZE": str(
                     framework_video_runtime["video_cache_size"]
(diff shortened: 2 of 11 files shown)


ramanathan831:
TAO-side live forced-replay verification on the unchanged official Brev CLI v0.6.334:

- Baseline unguarded submit: the first remote execution created container `58e81407d702`; Brev replayed the command, the second `docker run` hit the stable-name conflict, and the submit exited 1.
- Guarded submit from this PR: the first remote execution created container `072be85f6764`; Brev replayed the command, `docker inspect '$JOB_ID'` recognized the existing launch and returned `already submitted`; the submit exited 0.
- Inspection found exactly one container for each run. Both temporary containers were removed and verified absent.
- Focused contract regressions rerun after the live proof: 2 passed.

The separate Brev CLI PR was closed; retry safety is being handled at the TAO Brev submit-contract boundary.

skills/models/tao-finetune-video-clip/references/spec_template_export.yaml
```diff
@@ -26,7 +26,7 @@ export:
   input_height: 224
   input_width: 224
   input_channel: 3
-  batch_size: 1
+  batch_size: -1
```
vpraveen-nv: Collision: #203 modifies this same file — the same `batch_size: -1` change plus an `encoder_type: combined` flip that this PR does not carry. Whichever merges second conflicts. Since #203 owns the full Video-CLIP deploy story, suggest dropping this hunk here and letting #203 carry the template (6670029 is tracked by both PRs, so coverage is preserved).

skills/models/tao-finetune-cosmos-reason/scripts/framework_checkpoint_action.py
```diff
@@ -84,10 +84,25 @@ def infer_config_file(checkpoint: Path) -> Path:
...
 def _normalized_source(value: str) -> str:
+    if value.startswith("hf_model://"):
+        return value.removeprefix("hf_model://").strip("/")
+    if value.startswith("hf://"):
+        return value.removeprefix("hf://").removeprefix("models/").strip("/")
     candidate = Path(value).expanduser()
     return str(candidate.resolve()) if candidate.exists() else value
 
 
+def _indexed_tensor_keys(root: Path) -> set[str] | None:
```
vpraveen-nv: NVBug 6670478's complaint was that the pre-action "never checks tensor keys, only metadata-file hashes" — this closes that for indexed checkpoints, but returns None (check skipped, silently) whenever either side lacks `model.safetensors.index.json`, i.e. exactly the single-file-safetensors case. Suggest either reading key sets from the safetensors header directly (the first 8 bytes give the JSON header length; no tensor data needs loading) as a fallback, or at minimum logging a visible warning that tensor-key verification was skipped, so the gap is loud rather than silent.

skills/models/tao-finetune-cosmos-reason/scripts/cosmos_workflow.py
```diff
@@ -3531,6 +3539,30 @@ def build_plan(
...
+            ["nvidia-smi", "--query-gpu=memory.total", "--format=csv,noheader,nounits"],
+            text=True,
+            capture_output=True,
+            check=False,
+        )
+        if detected.returncode:
+            raise WorkflowError(
+                "could not derive --gpus-per-node from the Docker host; specify it explicitly"
+            )
+        eligible = [
```
vpraveen-nv: The >=16 GiB eligibility filter is a sensible display-GPU heuristic, but `nvidia-smi` enumerates all physical GPUs regardless of `CUDA_VISIBLE_DEVICES`. On a shared host where the user is restricted to a subset, the derived `gpus_per_node` will exceed what the container can actually claim and the job launches overcommitted. Suggest intersecting with `CUDA_VISIBLE_DEVICES` when it is set (same convention check_tao_launch_preflight.py applies).

skills/models/tao-finetune-cosmos-reason/scripts/cosmos_common.py
```diff
@@ -222,6 +225,27 @@ def _record_media(record: Mapping[str, Any]) -> list[str]:
     return values
 
 
+def decode_media(path: Path) -> None:
```
vpraveen-nv, line 228: Scale note: `local_preflight` now ffprobes every video in both media manifests serially — one subprocess per clip. Fine for smoke packs; for a real dataset (thousands of clips) preflight becomes minutes of ffprobe spawns. Suggest a `--decode-sample N` escape hatch (probe N random clips + always the first/last) or a parallel map, keeping full-probe as the default for small manifests. Not blocking.

skills/models/tao-finetune-video-clip/references/spec_template_export.yaml
ramanathan831: Addressed in `4c3cad9`: I removed this template hunk and the corresponding assertion from #201. The branch now has no diff for `spec_template_export.yaml`; #203 remains the single owner of the complete Video-CLIP deploy change.

skills/models/tao-finetune-cosmos-reason/scripts/framework_checkpoint_action.py
ramanathan831: Addressed comprehensively in `4c3cad9`. Verification now reads actual safetensors headers without tensor payloads for single-file and indexed/sharded checkpoints. Indexed key maps are cross-checked against shard headers, paths must remain contained, duplicate/malformed/missing entries fail closed, and base/export key sets are always compared for local safetensors models. Added a single-file drift regression in addition to the indexed case; the relevant suite passes.

skills/models/tao-finetune-cosmos-reason/scripts/cosmos_workflow.py
ramanathan831: Addressed in `4c3cad9`. Docker GPU inventory now includes physical index, UUID, and memory; `CUDA_VISIBLE_DEVICES` is resolved by physical index or unique GPU-UUID prefix before the >=16 GiB filter. The selected host IDs are sealed into the plan and the exact `device=...` request is used by both container preflight and final Docker launch, so the fix covers allocation as well as the derived count. Index/UUID, low-memory subset, and rendered-device regressions pass.

skills/models/tao-finetune-cosmos-reason/scripts/cosmos_common.py
ramanathan831: Addressed in `4c3cad9` with a bounded eight-worker ffprobe pool. Full decode validation remains fail-closed for every manifest clip; I intentionally avoided sampling because it would let corruption in an unsampled clip reach the job.

christinayyw:

**Latest run:** [tao-sb-OSS-develop #346](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/346/) — `83af5f67` @ 2026-08-27 22:59 UTC

|---|---|---|---|---|
| `skills/applications/tao-run-deft-aoi-cosmos3` | `cosmos3-deft-bare-plan` | `claude` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/346/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-aoi-cosmos3-claude/skill-eval-report-cosmos3-deft-bare-plan.html) |
| `skills/applications/tao-run-deft-aoi-cosmos3` | `cosmos3-deft-bare-plan` | `codex` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/346/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-aoi-cosmos3-codex/skill-eval-report-cosmos3-deft-bare-plan.html) |
| `skills/applications/tao-run-deft-aoi-cosmos3` | `cosmos3-deft-loop-ag-mining` | `claude` | ✅ Pass (⚠️ w/ self-correction) | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/346/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-aoi-cosmos3-claude/skill-eval-report-cosmos3-deft-loop-ag-mining.html) |
| `skills/applications/tao-run-deft-aoi-cosmos3` | `cosmos3-deft-loop-ag-mining` | `codex` | ✅ Pass (⚠️ w/ self-correction) | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/346/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-aoi-cosmos3-codex/skill-eval-report-cosmos3-deft-loop-ag-mining.html) |
| `skills/models/tao-finetune-video-clip` | — | `claude` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/346/artifact/jenkins-eval-reports/skills-models-tao-finetune-video-clip-claude/skill-eval-report.html) |
| `skills/models/tao-finetune-video-clip` | — | `codex` | ✅ Pass (⚠️ w/ self-correction) | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/346/artifact/jenkins-eval-reports/skills-models-tao-finetune-video-clip-codex/skill-eval-report.html) |


vpraveen-nv (approved):
Approving: all four review threads verified addressed against the code at head `83af5f67`.

- **spec_template_export.yaml** — dropped; the file is absent from the changed-file set, so the collision with #203 is gone.
- **framework_checkpoint_action.py** — `_safetensors_file_tensor_keys()` parses the 8-byte header length + JSON header without loading tensors; `_safetensors_tensor_keys()` falls back to `root.glob("*.safetensors")`, cross-checks index keys against shard headers, and rejects duplicates and path escapes. `verify_export` now raises `WorkflowError` instead of skipping — fails closed, which was the ask.
- **cosmos_workflow.py** — `_docker_visible_gpu_inventory(cuda_visibility)` resolves each `CUDA_VISIBLE_DEVICES` token by index or unique `GPU-` UUID prefix before the 16 GiB filter, and the sealed indices render as `--gpus device=<ids>` at both the preflight and final launch sites.
- **cosmos_common.py** — bounded `ThreadPoolExecutor(max_workers=min(8, len(media_paths)))` over `decode_media`. Sampling was deliberately declined to keep validation fail-closed for every clip; that is a reasonable call on a non-blocking item.

Minor note for the record: the reply on the checkpoint thread cites commit `4c3cad9`, which isn't in this PR's commit list — the branch was amended to `83af5f67`. The content is present at head either way.

PR #175, opened 2026-08-24
[TAO-0000][Bugfix] NVBug 6641997, 6641998: DEFT script exec bits and flattened-install test imports

xiaosean:
NVBugs: [6641997](https://nvbugspro.nvidia.com/bug/6641997), [6641998](https://nvbugspro.nvidia.com/bug/6641998)

## Summary

- **NVBug 6641997** — `deft_exec.py`, `deft_context.py`, `finalize_run.py` shipped mode 100644 while SKILL.md invokes them directly (`Permission denied`, shell 126). All three are now 100755 — in BOTH variants (the aoi copies had the same defect). A new regression test scans DEFT application docs for directly-invoked scripts and asserts filesystem + git-index exec bits. Widening the platform-level guard (`test_platform_contract.py`) is left to its owner as a follow-up.
- **NVBug 6641998** — the Cosmos3 bundled test suite failed to import under a flattened `.claude/skills/` plugin install (fragile parent-directory counting). Companion model/data skills are now resolved from either the canonical `skills/{models,data}/` tree or flattened sibling directories; the regression test runs the full suite from a temporary flattened layout.

## Validation

- AOI 29 + Cosmos3 29 tests pass (incl. two new packaging tests); all six direct `--help` probes pass; flattened-layout discovery runs all 27 Cosmos tests.
- App-layer only (8 files); premise verified on `origin/main` (mode 100644, status 126) before fixing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #209, opened 2026-08-28
[TAO-0000][Bugfix] NVBug 6641997: make documented script invocations independent of the exec bit

xiaosean:
Fixes: https://nvbugspro.nvidia.com/bug/6641997

**NVBug 6641997** — Control scripts shipped 0644 caused Permission denied when invoked directly per SKILL.md. PR #175 fixed the repo modes (0755), but the plugin install channel strips executable bits when copying into the plugin cache (Claude Code upstream issue #38686, a repeat regression), so QA installing via `claude plugin install` still reproduces the failure.

Root fix: make every documented invocation independent of the exec bit.
- Both variants (`tao-run-deft-aoi`, `tao-run-deft-aoi-cosmos3`): all documented direct script invocations in SKILL.md / references now use interpreter form — `"$PYTHON" scripts/foo.py` (via the existing `scripts/deft_python.sh` pattern) and `bash scripts/bar.sh`. Semantics unchanged.
- Regression test upgraded: simulates a flattened, exec-bit-stripped install and asserts the documented invocation forms still work (mode assertions in the repo are kept as defense in depth).

## Validation
- Test suites: 38/38 and 53/53 passed (both variants); `validate-skills.sh` passed; `git diff --check` clean.
- Single signed commit, rebased on latest main.


skills/applications/tao-run-deft-aoi-cosmos3/tests/test_execbit_packaging.py
```diff
@@ -26,12 +30,27 @@
...
+    r"scripts/(?:[A-Za-z0-9_.-]+/)*[A-Za-z0-9_.-]+\.(?:py|sh))(?=[ `])",
+    re.IGNORECASE,
+)
 
 
 def _documented_direct_control_scripts() -> dict[pathlib.Path, set[pathlib.Path]]:
     """Find DEFT control programs that app docs instruct agents to execute."""
     found: dict[pathlib.Path, set[pathlib.Path]] = {}
-    for skill_root in sorted(APPLICATIONS_ROOT.glob("tao-run-deft-*")):
+    for name in SKILL_NAMES:
```
vpraveen-nv, line 52: This replaces `APPLICATIONS_ROOT.glob("tao-run-deft-*")` with a hardcoded two-name tuple, which narrows the **pre-existing** `test_control_scripts_remain_executable_in_repo` as well — `tao-run-deft-iaa`, `tao-run-deft-object-detection` and `tao-run-deft-cr-its-mining` were covered by that mode assertion on `main` and no longer are.

Running the new `_documented_bare_invocations` against those three siblings reports **49, 66 and 5** bare invocations respectively (all still the `"$SKILL_ROOT/scripts/deft_python.sh" ... foo.py` form), so they reproduce NVBug 6641997 identically under a plugin install that strips exec bits. Please either keep the glob and convert the siblings, or keep the glob for the mode assertion and add an explicit commented allow-list (with a follow-up bug number) for the not-yet-converted skills. As written, a reviewer reading the green suite would conclude the whole `tao-run-deft-*` family is fixed.

skills/applications/tao-run-deft-aoi/references/prepare-for-inference.md
```diff
@@ -12,7 +12,7 @@ downstream inference skills can consume the trained checkpoint without reading
 | `best_model_inference_spec.yaml` | Ready-to-run TAO inference spec. The executable artifact. |
 
 Both are written by `scripts/prepare_inference_spec.py`. At ordinary loop end,
-use `scripts/finalize_run.py`; it writes these first and refuses the terminal
+use `"$PYTHON" scripts/finalize_run.py`; it writes these first and refuses the terminal
```
vpraveen-nv, line 15: `$PYTHON` is used here but never assigned anywhere in this file, so an agent that reads only this reference and pastes the command into a fresh Bash call gets `: command not found` instead of the old `Permission denied` — a different failure, not a fixed one.

Same holds for `tao-run-deft-aoi/references/air-gap.md` (1 use, 0 assignments) and `tao-run-deft-aoi-cosmos3/references/{REPORT_RENDERING,pipeline-and-state,aoi-annotation}.md`. Since these docs are loaded independently of SKILL.md, each file referencing `"$PYTHON"` should either show the `PYTHON=$(bash <skill_root>/scripts/deft_python.sh)` line inline, or open with a one-line pointer to `references/scripts-and-agents.md`. A cheap guard: extend the new test to assert every doc containing `"$PYTHON"` either assigns it or carries that pointer.

skills/applications/tao-run-deft-aoi/references/scripts-and-agents.md
```diff
@@ -2,7 +2,8 @@
 
 ## Using Bundled Scripts
 
-Run every bundled script through `scripts/deft_python.sh`. It selects
+Set `PYTHON=$(bash scripts/deft_python.sh)` before every bundled-script
```
vpraveen-nv, line 5: `deft_python.sh` exits 2 with a stderr diagnostic when no candidate interpreter satisfies the import probe, but `PYTHON=$(bash ...)` in a non-`set -e` agent shell just leaves `PYTHON` empty and execution continues to `"$PYTHON" scripts/foo.py`, which fails as `command not found` — losing the actionable "no installed Python provides pandas,numpy,..." signal. The old single-command form propagated exit 2 directly.

Recommend documenting the guarded form once here and in `preflight.md`: `PYTHON=$(bash scripts/deft_python.sh) || exit 1` followed by `: "${PYTHON:?deft_python.sh selected no interpreter}"`. This matters most in air-gap mode, where the documented behaviour is a hard stop rather than a fallback.

skills/applications/tao-run-deft-aoi-cosmos3/tests/test_execbit_packaging.py
```diff
@@ -26,12 +30,27 @@
 SCRIPT_REFERENCE_RE = re.compile(
     r"(?<![A-Za-z0-9_.-])scripts/(?P<name>[A-Za-z0-9_.-]+\.(?:py|sh))"
 )
+SHELL_CONTINUATION_RE = re.compile(r"\\\n[ \t]*")
+BARE_SCRIPT_EXEC_RE = re.compile(
```
vpraveen-nv, line 34: I probed this detector against plausible future additions and it misses several realistic un-converted forms: `docker run $IMG scripts/foo.py`, `time scripts/foo.py`, `exec scripts/foo.py`, `env X=1 scripts/foo.py` and `if scripts/foo.py; then` — none of the leading alternatives (`^`, `&&`, `||`, `;`, `$(`, `bash -lc "`) match those positions.

`INLINE_BARE_SCRIPT_EXEC_RE` is also gated on a fixed verb whitelist, so ``then `scripts/deft_exec.py --state ...` `` slips through while the near-identical ``run `scripts/deft_exec.py ...` `` is caught. Adding `time|exec|env|xargs|sudo` and `if|while|until` to the prefix alternation, plus a token-position match after a `docker run ... <image>` tail, would close the highest-risk gaps without adding false positives on the many legitimate prose mentions.

skills/applications/tao-run-deft-aoi-cosmos3/tests/test_execbit_packaging.py
```diff
@@ -51,6 +70,28 @@ def _documented_direct_control_scripts() -> dict[pathlib.Path, set[pathlib.Path]
...
+        for pattern in (BARE_SCRIPT_EXEC_RE, INLINE_BARE_SCRIPT_EXEC_RE):
+            for match in pattern.finditer(text):
+                line = text.count("\n", 0, match.start()) + 1
+                found.append(
+                    f"{doc.relative_to(skill_root)}:{line}: {match.group('command')}"
+                )
+    return found
+
+
+def _strip_file_exec_bits(root: pathlib.Path) -> None:
```
vpraveen-nv, line 89: The exec-bit-stripped install simulation is mostly decorative for the assertion it guards: `_documented_bare_invocations` is a pure regex scan over Markdown text, so the `copytree` + `chmod 0o644` has no effect on its result — it would return identically if run against `APPLICATIONS_ROOT` in place. Only the `bash deft_python.sh` subprocess check is actually mode-sensitive, and that exercises the launcher rather than any documented command.

Consider making the simulation earn its keep: extract one representative documented command per skill (e.g. the `commit_stage.py` block) from the stripped copy and execute it with `DEFT_PYTHON` pointed at a stub, asserting a non-127 exit. Otherwise the tempdir machinery invites a future reader to trust coverage the test doesn't have.

vpraveen-nv (commented):
minor comments.

vpraveen-nv (approved):

christinayyw:

**Latest run:** [tao-sb-OSS-develop #401](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/401/) — `0d793d97` @ 2026-08-29 08:06 UTC

|---|---|---|---|---|
| `skills/applications/tao-run-deft-aoi` | — | `claude` | ✅ Pass (⚠️ w/ self-correction) | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/401/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-aoi-claude/skill-eval-report.html) |
| `skills/applications/tao-run-deft-aoi` | — | `codex` | ❌ Fail | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/401/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-aoi-codex/skill-eval-report.html) |
| `skills/applications/tao-run-deft-aoi-cosmos3` | `cosmos3-deft-bare-plan` | `claude` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/401/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-aoi-cosmos3-claude/skill-eval-report-cosmos3-deft-bare-plan.html) |
| `skills/applications/tao-run-deft-aoi-cosmos3` | `cosmos3-deft-bare-plan` | `codex` | ✅ Pass | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/401/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-aoi-cosmos3-codex/skill-eval-report-cosmos3-deft-bare-plan.html) |
| `skills/applications/tao-run-deft-aoi-cosmos3` | `cosmos3-deft-loop-ag-mining` | `claude` | ❓ Error | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/401/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-aoi-cosmos3-claude/skill-eval-report-cosmos3-deft-loop-ag-mining.html) |
| `skills/applications/tao-run-deft-aoi-cosmos3` | `cosmos3-deft-loop-ag-mining` | `codex` | ✅ Pass (⚠️ w/ self-correction) | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/401/artifact/jenkins-eval-reports/skills-applications-tao-run-deft-aoi-cosmos3-codex/skill-eval-report-cosmos3-deft-loop-ag-mining.html) |

PR #194, opened 2026-08-25
[TAO-6655434][Bugfix] Rename DEFT workflow from IAA to PAS

ramanathan831:
## Summary

- rename the customer-facing DEFT workflow to People Attribute Search (PAS)
- rename the skill, package, scripts, state/config keys, tests, marketplace registration, and documentation
- consume the evaluator-owned `nvidia_pas_metrics*.csv` artifact names emitted by TAO 7.2
- add a regression guard for the canonical PAS workflow and package surface

## Validation

- `python -m pytest -q scripts/tests/test_pas_deft_container.py skills/applications/tao-run-deft-aoi/tests/test_runtime_hygiene.py` — 20 passed
- `python -m pytest -q scripts/tests` — 362 passed, 2 skipped
- `python scripts/stamp_versions.py --check --strict-strays` — passed with 0 stray warnings
- `scripts/validate-skills.sh --all` — passed with the two pre-existing NV-Tesseract container-image warnings
- Python, JSON, YAML, and shell parsing — passed
- obsolete workflow/package identifier scan — passed
- refreshed-plugin import and capability discovery — passed
- `git diff --check` — passed

Fixes NVBug 6655434.

PR #87, opened 2026-08-05
[TAO-0000][Feature] Migrate standalone AutoML skills

ramanathan831:
## Summary

Port the standalone AutoML skill implementation and its deterministic validation coverage onto current GitHub main.

This is the GitHub-main replacement for work developed before the GitLab repositories were discontinued.

## Migration contract

- Base: `main`
- Source branch: `dev/ram/automl-skill-migration-main`
- Branch prefix: `dev/ram/`
- Repository history, not an image or generated runtime patch, is the source of truth.
- No credentials or user-specific runtime paths are included.

PR #210, opened 2026-08-28
[TAO-0000][Chore] Trim tao-run-automl SKILL.md under the 20000-char signer cap

xiaosean:
Unblocks repo-wide CI: `validate-skills.sh` check 3b currently fails every PR because `skills/applications/tao-run-automl/SKILL.md` is 20,040 chars, over the 20,000-char signer cap (observed on PR #209's blossom-ci and reproducible on plain main).

- Six meaning-preserving condensations (wording only, no semantic change): 19,966 chars, back under the cap.
- Longer-term the AutoML skill owner may want to move detail into `references/` per the validator's guidance — this is the minimal hotfix to turn main green again.

## Validation
- `validate-skills.sh` passes locally on the branch (check 3b no longer errors); `git diff --check` clean; single signed commit.

### DIFF
--- a/skills/applications/tao-run-automl/SKILL.md
+++ b/skills/applications/tao-run-automl/SKILL.md
@@ -124,7 +124,7 @@ Collect these before runner construction:
 |---|---|
 | `model_skill` | Resolved model skill directory under `skills/models/`. Resolve user aliases such as `network_arch` to the packaged skill directory first. |
 | `network_arch` | Read from the resolved model skill metadata. |
-| `action` | Action to optimize — `train`, `evaluate`, `inference`, `distill`, `prune`, or `quantize` with a packaged schema/template. |
+| `action` | Action to optimize, usually `train`, `evaluate`, `inference`, `distill`, `prune`, or `quantize` with a packaged schema/template. |
 | `platform` | One of the supported TAO platform skills. |
 | `train_dataset` / `eval_dataset` / action inputs | Use model-specific spec keys and layout. Non-train actions may also need parent/teacher checkpoints, calibration data, or pruned artifacts. |
 | `results_root` | Local, Lustre, or S3 path appropriate for the platform. |
@@ -233,7 +233,7 @@ adjustment in `result["history"][i]["adjustments"]`.
 | `hyperband`, `asha` | Many configs, cheap early rungs; ASHA is parallel-friendly. | `max_epochs`, `reduction_factor`, optional `max_concurrent` |
 | `bohb`, `dehb` | Mixed Bayesian/evolutionary search with multi-fidelity budgets. | same rung budget fields as Hyperband |
 | `pbt` | Long training where schedules should mutate during training. | population and generation budget |
-| `llm`, `hybrid`, `autoresearch` | User explicitly wants LLM-guided search with a configured endpoint. | LLM endpoint config plus budget |
+| `llm`, `hybrid`, `autoresearch` | User explicitly wants LLM-guided search and an endpoint is configured. | LLM endpoint config plus budget |
 
 For `evaluate` or `inference`, default to Bayesian/BFBO-style search over the
 selected action's prompt, decoding, preprocessing, or runtime config knobs.
--- a/skills/applications/tao-run-automl/SKILL.md
+++ b/skills/applications/tao-run-automl/SKILL.md
@@ -25,7 +25,7 @@ tags:
 
 # TAO AutoML
 
-> **Standalone install?** If this session was not initialized by the TAO skill bank plugin, run the `tao-setup` skill first (host preflight, credentials, cross-skill discovery).
+> **Standalone install?** If this session was not initialized by the TAO skill bank plugin, run `tao-setup` first (host preflight, credentials, cross-skill discovery).
 
 Run automated hyperparameter optimization for a TAO model by combining:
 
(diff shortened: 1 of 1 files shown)


christinayyw:

**Latest run:** [tao-sb-OSS-develop #370](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/370/) — `ecd5ed11` @ 2026-08-28 15:42 UTC

|---|---|---|---|---|
| `skills/applications/tao-run-automl` | — | `claude` | ✅ Pass (⚠️ w/ self-correction) | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/370/artifact/jenkins-eval-reports/skills-applications-tao-run-automl-claude/skill-eval-report.html) |
| `skills/applications/tao-run-automl` | — | `codex` | ❌ Fail | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/370/artifact/jenkins-eval-reports/skills-applications-tao-run-automl-codex/skill-eval-report.html) |


skills/applications/tao-run-automl/SKILL.md
```diff
@@ -122,11 +122,11 @@ Collect these before runner construction:
 
 | Input | Requirement |
 |---|---|
-| `model_skill` | Resolved model skill directory under `skills/models/`. Accept user aliases such as `network_arch` only after resolving them to the packaged skill directory. |
+| `model_skill` | Resolved model skill directory under `skills/models/`. Resolve user aliases such as `network_arch` to the packaged skill directory first. |
 | `network_arch` | Read from the resolved model skill metadata. |
-| `action` | Action to optimize, usually `train`, `evaluate`, `inference`, `distill`, `prune`, or `quantize` when that action has a packaged schema/template. |
+| `action` | Action to optimize — `train`, `evaluate`, `inference`, `distill`, `prune`, or `quantize` with a packaged schema/template. |
```
vpraveen-nv: This rewrite drops the word **"usually"**, which changes meaning rather than preserving it. The original — "Action to optimize, usually `train`, `evaluate`, `inference`, `distill`, `prune`, or `quantize` when that action has a packaged schema/template" — presents an illustrative list gated on a condition. The new phrasing reads as a closed enumeration, so an agent following the skill may refuse a legitimate action (e.g. `export`, `gen_trt_engine`) that does have a packaged schema/template.

Restoring "usually" costs 8 chars and there are 46 chars of headroom, so the cap isn't a reason to drop it. Suggested: ``Action to optimize, usually `train`, `evaluate`, `inference`, `distill`, `prune`, or `quantize` with a packaged schema/template.``

skills/applications/tao-run-automl/SKILL.md
```diff
@@ -25,7 +25,7 @@ tags:
 
 # TAO AutoML
 
-> **Standalone install?** If this session was not initialized by the TAO skill bank plugin, run the `tao-setup` skill first (host preflight, credentials, cross-skill discovery).
+> **Standalone install?** If this session was not initialized by the TAO skill bank plugin, run `tao-setup` first (host preflight, credentials, cross-skill discovery).
```
vpraveen-nv, line 28: Confirmed the fix works: at head `ecd5ed1` the file is **19,954 chars** against the 20,000 ceiling in `validate-skills.sh` check 3b, down from 20,040 on main. (The PR description quotes 19,966 bytes; check 3b uses `len(f.read())`, so chars is the number that matters.)

But 46 chars of margin isn't durable, and check 3b only runs under `scripts/validate-skills.sh`, which **no always-on workflow invokes** — `static-tests.yml` runs pre-commit only, and blossom-ci is comment-gated `/build`. There's also no test under `scripts/tests/` asserting the ceiling. This is repo-wide: `tao-train-mask2former/SKILL.md` is at 19,977 chars (23 to spare) and 10 other SKILL.md files sit within 600 chars of the cap, so `main` will go red again on the next one-paragraph edit. Worth pairing this hotfix with a cheap pre-commit hook or a `scripts/tests/` unit test running just the 3b size check on changed SKILL.md files.

skills/applications/tao-run-automl/SKILL.md
```diff
@@ -230,10 +230,10 @@ adjustment in `result["history"][i]["adjustments"]`.
 | Algorithm | Good fit | Required knobs |
 |---|---|---|
 | `bayesian` | Default for small/medium budgets and few parameters. | `num_recommendations`, metric, direction |
-| `hyperband`, `asha` | Many configs with cheap early rungs; ASHA supports parallelism. | `max_epochs`, `reduction_factor`, optional `max_concurrent` |
+| `hyperband`, `asha` | Many configs, cheap early rungs; ASHA is parallel-friendly. | `max_epochs`, `reduction_factor`, optional `max_concurrent` |
 | `bohb`, `dehb` | Mixed Bayesian/evolutionary search with multi-fidelity budgets. | same rung budget fields as Hyperband |
 | `pbt` | Long training where schedules should mutate during training. | population and generation budget |
-| `llm`, `hybrid`, `autoresearch` | User explicitly wants LLM-guided search and has an endpoint configured. | LLM endpoint config plus budget |
+| `llm`, `hybrid`, `autoresearch` | User explicitly wants LLM-guided search with a configured endpoint. | LLM endpoint config plus budget |
```
vpraveen-nv: The original stated two independent preconditions joined by "and": the user explicitly wants LLM-guided search, **and** an endpoint is configured. Folding the second into a prepositional phrase makes it read as one compound preference rather than a gate the agent must verify before selecting `llm`/`hybrid`/`autoresearch` — a path that fails at runtime with no endpoint. Keeping "and has an endpoint configured" costs 4 chars over the current wording and fits in the headroom.

Separately, note this trim is likely temporary: draft PR #87 brings this SKILL.md to 18,391 chars and independently clears the cap. #87's head still contains all six pre-trim lines verbatim, so the hunks don't textually collide, but whoever rebases #87 will probably take that branch's file wholesale and silently revert these six condensations — which is fine at 18,391 chars. Merging this as the stopgap is right; just don't expect the wording to survive #87.

skills/applications/tao-run-automl/SKILL.md
xiaosean: Restored "usually" in 7af640b — agreed the absolute phrasing changed meaning.

skills/applications/tao-run-automl/SKILL.md
xiaosean: Restored the two independent preconditions ("...and an endpoint is configured") in 7af640b. File is 19,964 chars after both restorations — still under the cap.

vpraveen-nv:
Heads-up: I think `main` may have overtaken this one.

At this PR's merge-base (`d2ff1fa`) `SKILL.md` was **20,040** chars, and these rewordings bring it to **19,964** — a real fix at the time. But `main` has since been trimmed independently (via the IAA→PAS rename in #194) and now sits at **19,954**, so applying this on top of current `main` would be **+10** chars rather than a reduction. That's likely also why it's showing conflicts.

Is this still worth rebasing from your side, or has the cap issue been overtaken by events? Happy to be wrong if you're keeping it for the prose improvements themselves — several of the rewordings do read better.

xiaosean:
Closing — superseded: main was independently trimmed below the cap by the IAA→PAS rename (#194), so this hotfix is no longer needed (and would now be +10 chars on top of current main, as Varun noted). Thanks for the verification and the wording catches!

PR #217, opened 2026-08-29
[TAO-0000][Bugfix] NVBug 6662975: document VirtualEnvSDK construction and its work_dir default

vpraveen-nv:
## NVBug 6662975

No skill documented how to construct `VirtualEnvSDK`. Three skills forward-referenced each other and the loop never closed:

- the NV-Tesseract model refs said *"use `tao-run-automl` to create the VirtualEnvSDK-backed AutoMLRunner"*
- `tao-run-automl`'s **Runner Construction** section never defined `sdk`
- its preflight reference said *"Construct the SDK with no arguments"* — actively wrong for this SDK

The consequence is not just confusion. The real signature is `VirtualEnvSDK(venv_path, work_dir=None, state_file=None)`, and when `work_dir` is omitted the SDK resolves it to `$TAO_SDK_STATE_DIR/virtualenv` or `~/.tao_sdk/virtualenv` and creates `<work_dir>/jobs/` on construction. Every trial's checkpoints then land in the home directory — the reporter measured ~1.39 GB per NV-Tesseract Forecasting trial.

## Change

- **`tao-run-automl/SKILL.md`**, Runner Construction: define `sdk`, give the venv constructor with both kwargs, state the home-directory default as a caution, point at the reference.
- **`references/automl-runner-configuration.md`**: new "VirtualEnvSDK (containerless venv runs)" section — full signature, caution, job layout, declared-output routing rule.
- **`references/automl-preflight-concepts.md`**: add `VirtualEnvSDK` to the SDK table; scope the "no arguments" instruction to the credential-reading SDKs and name `VirtualEnvSDK` as the exception.
- **both NV-Tesseract `references/automl.md`**: replace the dead forward-reference with a file-and-section link plus the constructor shape.

The two model-ref files are the closing half of the loop, not incidental scope — fixing only the `tao-run-automl` side would leave the dead forward-references in place.

## The check-4 trap

The full construction can only live under `skills/applications/tao-run-automl/` because `validate-skills.sh` check 4 bans the literal `tao_sdk` everywhere else. That is why the model refs say "a hidden state directory under `$HOME`" rather than pasting `~/.tao_sdk/virtualenv` — that literal would fail CI in a model skill. The new test pins this, so a future editor gets a test failure instead of a CI failure.

## Char budget

`SKILL.md` was 46 chars from the 20,000-char signer cap, so three **restatements** were removed to make room:

1. the baseline-eval directive already stated verbatim in `## Automatic Baseline Eval Job`
2. one of two consecutive pointers to the same reference file (merged into one naming all four topics)
3. a rhetorical sentence whose directive is carried by the sentence before it

No instruction is lost. Net **19,954 → 19,882** chars, raising headroom from 46 to 118.

## Verification

`scripts/validate-skills.sh` passes. 8 new tests, 5 of which are confirmed to **fail without the fix**. The 1.4 GB/trial figure is the reporter's measurement, not re-measured here.


skills/applications/tao-run-automl/references/automl-runner-configuration.md
```diff
@@ -73,6 +74,59 @@ result = runner.run(
...
+Declared outputs from `references/skill_info.yaml` are rewritten into that
+per-job results directory, so `train.output_dir` becomes
+`<work_dir>/jobs/<job-id>/results/train_output_dir`. Resolve it at runtime with
+`sdk.get_job_results_dir(job_id)`; do not precompute it.
+
+Do not pass `image=` for `python_script` execution - the venv is the runtime.
+Verify construction during preflight:
+
+```bash
+python -c "from tao_sdk.platforms.virtualenv import VirtualEnvSDK; print('OK')"
```
ramanathan831: `P2` This command only verifies that `VirtualEnvSDK` imports; it does not verify construction as the preceding sentence claims. It can therefore report `OK` even when the selected venv is invalid and actual runner construction will fail.

Reproduced on this PR head (`05b06863accef9944e5a10162c2776b9b3d5276c`):

```bash
repro_root=$(mktemp -d /tmp/pr217-construction-repro.XXXXXX)

# The command currently documented by this PR.
python -c 'from tao_sdk.platforms.virtualenv import VirtualEnvSDK; print("OK")'
printf 'documented_preflight_exit=%s\n' "$?"

# Actual construction using the selected venv/work directory.
PR217_MISSING_VENV="$repro_root/missing-model-venv" \
PR217_WORK_DIR="$repro_root/automl-jobs" \
python - <<'PY'
import os
from tao_sdk.platforms.virtualenv import VirtualEnvSDK

VirtualEnvSDK(
    venv_path=os.environ["PR217_MISSING_VENV"],
    work_dir=os.environ["PR217_WORK_DIR"],
)
print("CONSTRUCTED")
PY
printf 'actual_constructor_exit=%s\n' "$?"
```

Observed:

```text
OK
documented_preflight_exit=0
ValueError: Virtual environment does not exist: /tmp/pr217-construction-repro.../missing-model-venv
actual_constructor_exit=1
```

`python -m pytest -q scripts/tests/test_automl_virtualenv_sdk_docs.py` still reports `8 passed` because those tests assert documentation strings but never construct the SDK.

Please make the preflight instantiate `VirtualEnvSDK` with the selected `venv_path` and explicit `work_dir` (as the example above this section requires), and add coverage that would fail when the documented check is import-only.

ramanathan831 (approved):

skills/applications/tao-run-automl/references/automl-runner-configuration.md
vpraveen-nv: Fixed in `0d459df`. You were right on both counts — the command was import-only and the tests couldn't tell.

**The preflight now constructs.** It instantiates `VirtualEnvSDK` with the same `venv_path`/`work_dir` the runner will use, and the surrounding prose says why importing is not enough. I also documented the failure signature (`ValueError: Virtual environment does not exist: <venv_path>`) and noted that construction creates `<work_dir>/jobs/` — a second reason to pass `work_dir` explicitly.

**Coverage that fails on import-only.** Three new tests:

- `test_preflight_check_constructs_rather_than_imports` — string guard: asserts the block contains `VirtualEnvSDK(`, `venv_path=`, `work_dir=`.
- `test_documented_preflight_fails_on_a_missing_venv` — extracts the python from the documented heredoc and **executes it** against a stub carrying the real constructor's contract; asserts non-zero exit and the `ValueError` text.
- `test_documented_preflight_passes_on_a_valid_venv` — same command, valid venv: exit 0, stdout `OK`, and `<work_dir>/jobs/` created.

Verified in both directions. Against the fixed doc: **11 passed**. Reverting only the doc back to the import-only one-liner and re-running, all three fail with diagnostic assertions:

```text
test_preflight_check_constructs_rather_than_imports -> preflight never constructs the SDK
test_documented_preflight_fails_on_a_missing_venv   -> preflight is not a python heredoc; an import-only one-liner cannot validate venv_path
test_documented_preflight_passes_on_a_valid_venv    -> preflight is not a python heredoc; an import-only one-liner cannot validate venv_path
```

`scripts/validate-skills.sh` passes; `SKILL.md` unchanged at 19,894 chars.

christinayyw:

**Latest run:** [tao-sb-OSS-develop #418](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/418/) — `0d459dff` @ 2026-09-01 01:59 UTC

|---|---|---|---|---|
| `skills/applications/tao-run-automl` | — | `claude` | ✅ Pass (⚠️ w/ self-correction) | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/418/artifact/jenkins-eval-reports/skills-applications-tao-run-automl-claude/skill-eval-report.html) |
| `skills/applications/tao-run-automl` | — | `codex` | ❌ Fail | [Latest Eval Report](https://prod.blsm.nvidia.com/metropolis-iva-tlt-jenkins/job/tao-sb-OSS-develop/418/artifact/jenkins-eval-reports/skills-applications-tao-run-automl-codex/skill-eval-report.html) |

PR #226, opened 2026-09-02
[TAO-2501][Bugfix] kpi: score the way the reference pipeline actually scores

arihantj-nv:
## Why

`num_recall_points` and `ignore_sqwidth` were pinned at `11` and `40` — in the DEFT OD overlay and in the `tao-analyze-detection-kpi` asset — each with a comment naming the reference ITS pipeline as the authority:

> `num_recall_points: 11` … *"which is what the reference ITS pipeline these metrics are compared against uses"*
> `ignore_sqwidth: 40` … *"matching the reference pipeline"*

All seven reference specs use `101` and `0`:

```
SanTomas_HL_kpi_analyze.yaml                     recall=101 sqwidth=0
SanTomas_v3_HL_kpi_analyze.yaml                  recall=101 sqwidth=0
SanTomas_v3_held_out_HL_kpi_analyze.yaml         recall=101 sqwidth=0
SanTomas_v3_held_out_no_roi_HL_kpi_analyze.yaml  recall=101 sqwidth=0
SanTomas_v3_no_pic_HL_kpi_analyze.yaml           recall=101 sqwidth=0
SanTomas_v3_no_roi_HL_kpi_analyze.yaml           recall=101 sqwidth=0
SanTomas_v3_no_roi_no_pic_HL_kpi_analyze.yaml    recall=101 sqwidth=0
```

So the pins have contradicted their own stated authority since the skill was added.

## What

- `num_recall_points` → `101` (COCO-style interpolation), `ignore_sqwidth` → `0` (score every box).
- Applied in both places that hold the values — the workflow overlay and the leaf skill's asset — so the two still agree, which the workflow doc asserts.
- Corrected the four places the values are restated in prose, including the two comments that cited the reference incorrectly.

## Impact

**This changes every reported AP.** Numbers measured before this commit are not comparable to numbers measured after it, including the baselines quoted in earlier run reports. The change makes them comparable to the reference pipeline, which is what they were always presented as.

### DIFF
--- a/skills/applications/tao-run-deft-object-detection/assets/overlays/kpi_analyze.yaml
+++ b/skills/applications/tao-run-deft-object-detection/assets/overlays/kpi_analyze.yaml
@@ -18,11 +18,14 @@ visualize.platform: local      # TAO: local. Pinned so a loop stage cannot acqui
                                # a wandb credential dependency.
 
 kpi.iou_threshold: 0.5         # TAO: 0.5
-kpi.num_recall_points: 11      # TAO: 11. 11-point interpolated AP (VOC); 101
-                               # selects COCO-style and shifts every reported AP.
-kpi.ignore_sqwidth: 40         # TAO: 0. Excludes boxes below this square-width
-                               # from scoring; 0 scores small boxes the reference
-                               # set excludes.
+kpi.num_recall_points: 101     # TAO: 11. 101-point interpolation, COCO-style, rather
+                               # than the 11-point VOC curve. Every reported AP moves
+                               # with this, so a run scored at 11 is not comparable to
+                               # one scored here.
+kpi.ignore_sqwidth: 0          # TAO: 0. Score every box, including small ones. Raising
+                               # it drops boxes below that square-width from scoring
+                               # entirely, which flatters AP by removing the hardest
+                               # detections rather than by detecting them.
 kpi.filter: false              # TAO: false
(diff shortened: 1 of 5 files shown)


ddevendran-nvidia (commented):
Could you add a testing section? i.e. how you tested this?

vpraveen-nv (commented):
Agree with the change. Two things:

1. One prose site missed: `scripts/apply_spec_overrides.py` docstring (~L24) still says "TAO's default for `kpi.ignore_sqwidth` is 0 where this workflow needs 40". Now false — swap the example for `num_recall_points` (11→101).
2. A run whose baseline was scored at 11/40 and is resumed after this lands gets a trend from two scoring regimes, and nothing on disk flags it. Not for this PR, but recording `num_recall_points`/`ignore_sqwidth` in `kpi_summary.json` and having the audit flag mixed phases would close it. I can file the ticket.

Also: the SanTomas reference specs aren't in this repo, so record where they live in the asset comment — the whole justification is "matches the reference".

arihantj-nv:
@ddevendran-nvidia added a Testing section. It rescores the 5k run's baseline and iter1 under both settings on the pinned DS image. The old settings reproduce the original numbers exactly. The new ones drop mAP by 0.074 at baseline, mostly from person (−0.248), and raise the loop's measured baseline→iter1 gain from +0.017 to +0.043.

ddevendran-nvidia (approved):

PR #245, opened 2026-09-10
[TAO-2529][Feature] Add a thin skill interface for the Data Services DINOv3 DEFT controller

nnagrajrao:
### What changes are proposed in this pull request?

Add a thin DINOv3 SSL DEFT skill that delegates execution to the packaged Data Services controller and documents the sealed contracts.

**DEFT for SSL**

This stack implements iterative data refinement around DINOv3 self-supervised training. Each round scores the immutable target population, selects weak acquisition targets, retrieves related samples from an immutable unlabeled source pool, deduplicates them, and appends file/archive locators to a cumulative training manifest. Every candidate starts from the original checkpoint—not the previous candidate's weights—and optional sealed evaluation can stop the loop. GRIT provides label-free representation-instability rankings; it is not calibrated uncertainty and the workflow does not guarantee an accuracy improvement.

**PR stack and review order (TAO-2529)**

- PyTorch: [#132](https://github.com/NVIDIA-TAO/tao-pytorch/pull/132) → [#133](https://github.com/NVIDIA-TAO/tao-pytorch/pull/133) → [#134](https://github.com/NVIDIA-TAO/tao-pytorch/pull/134)
- Data Services: [#38](https://github.com/NVIDIA-TAO/tao-data-services/pull/38) → [#39](https://github.com/NVIDIA-TAO/tao-data-services/pull/39) → [#40](https://github.com/NVIDIA-TAO/tao-data-services/pull/40)
- Core: [#38](https://github.com/NVIDIA-TAO/tao-core/pull/38)
- Skill Bank: [#245](https://github.com/NVIDIA-TAO/tao-skill-bank/pull/245) → [#246](https://github.com/NVIDIA-TAO/tao-skill-bank/pull/246)

Within a repository, arrows are Git base dependencies and should be reviewed and merged left to right. Cross-repository links are runtime/API dependencies. This increment is based on `main`, contains one signed-off commit, and is headed by `nnagrajrao/deft/01-controller`.

### Why are the changes needed?

Users need a discoverable TAO workflow without duplicating production orchestration in Skill Bank.

### Related issues

JIRA: [TAO-2529](https://jirasw.nvidia.com/browse/TAO-2529)

### Does this PR introduce _any_ user-facing change?

Adds one DINOv3-specific application skill.

### How was this patch tested?

The final clean committed snapshots were archived and tested inside the approved TAO 7.2 rc36 containers:

[...]

PR #246, opened 2026-09-10
[TAO-2529][CI] Add manual cross-repository DEFT container test harness

nnagrajrao:

Add a portable handoff that validates clean committed source snapshots and separately probes installed packages in pinned TAO images.

**DEFT for SSL**

This stack implements iterative data refinement around DINOv3 self-supervised training. Each round scores the immutable target population, selects weak acquisition targets, retrieves related samples from an immutable unlabeled source pool, deduplicates them, and appends file/archive locators to a cumulative training manifest. Every candidate starts from the original checkpoint—not the previous candidate's weights—and optional sealed evaluation can stop the loop. GRIT provides label-free representation-instability rankings; it is not calibrated uncertainty and the workflow does not guarantee an accuracy improvement.

**PR stack and review order (TAO-2529)**

- PyTorch: [#132](https://github.com/NVIDIA-TAO/tao-pytorch/pull/132) → [#133](https://github.com/NVIDIA-TAO/tao-pytorch/pull/133) → [#134](https://github.com/NVIDIA-TAO/tao-pytorch/pull/134)
- Data Services: [#38](https://github.com/NVIDIA-TAO/tao-data-services/pull/38) → [#39](https://github.com/NVIDIA-TAO/tao-data-services/pull/39) → [#40](https://github.com/NVIDIA-TAO/tao-data-services/pull/40)
- Core: [#38](https://github.com/NVIDIA-TAO/tao-core/pull/38)
- Skill Bank: [#245](https://github.com/NVIDIA-TAO/tao-skill-bank/pull/245) → [#246](https://github.com/NVIDIA-TAO/tao-skill-bank/pull/246)

Within a repository, arrows are Git base dependencies and should be reviewed and merged left to right. Cross-repository links are runtime/API dependencies. This increment is based on `nnagrajrao/deft/01-controller`, contains one signed-off commit, and is headed by `nnagrajrao/deft/02-container-ci`.

Reviewers need reproducible source compatibility evidence while keeping source correctness distinct from release-image readiness.

### Related issues

JIRA: [TAO-2529](https://jirasw.nvidia.com/browse/TAO-2529)

No production behavior change; this adds development validation tooling.

The final clean committed snapshots were archived and tested inside the approved TAO 7.2 rc36 containers:

- Final repository tips: Skill Bank `edef2e5d`, PyTorch `c22f3b80`, Data Services `7f2497e3`, Core `b3fd05f`.
- Source CI handoff: Skill Bank validation/hooks plus 27 tests; PyTorch hooks plus 323 passed and 4 expected weight-dependent skips; Data Services hooks plus 216 passed; Core hooks plus 19 passed.
- Broader QA: PyTorch 263 passed and 4 expected skips; Data Services 211 passed; Core 19 passed; Skill Bank 27 passed.
- Cross-repository dummy-data smoke: 5 passed, covering real image materialization/consumption, a DS LocalRunner process boundary into precomputed GRIT, exact embedding binding, schema equality, and DS/PyTorch attestation parity.
- Real Lightning smoke covered two local workers and two real nodes without torchrun.
- Independent adversarial panel re-review found no remaining source, contract, isolation, or test-coverage blocker.

Approved images were pinned to PyTorch `sha256:83eb5598411bf5a2b28f366e8861153e8da77f805edef0d200baa95bf3c5dc62` and DS `sha256:ab55b92d8c0eb570c29661dc8fe4a03dfe76cd5cc3a685ed77edcfc221c0f87a`. The source-free packaged probes fail only because rc36 predates the new PyTorch/DS modules and Core schema; rebuilding compatible release images remains a release-time requirement. No Dockerfile or dependency manifest changed.

Each GitLab integration branch is one signed-off commit over current `origin/main` (`0 1` ancestry), and its tree exactly matches the final GitHub stack tip. Production-scale throughput, real-data quality, full multi-node/FSDP training, and rebuilt packaged images remain outside this validation.

Generated-by: OpenAI Codex CLI

### Release note

```release-note
NONE
```

### Checklist

- [x] Commits are signed off per DCO.
- [x] Repository style, lint, and feature regression tests pass.
- [x] Tests and documentation cover the change.
- [x] No secrets, credentials, customer data, Dockerfiles, or dependency manifests are included.
- [ ] Published release images contain the complete feature stack — pending normal TAO image publication.

### Notes for reviewers

Review base: `nnagrajrao/deft/01-controller`; head: `nnagrajrao/deft/02-container-ci`. This PR is ready for review and is not a draft. The scope is DINOv3 SSL DEFT only; other models keep their existing behavior.


scripts/test_dinov3_deft_containers.py
```diff
@@ -0,0 +1,251 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Portable DEFT container checks using existing images, never Docker builds."""
```
vpraveen-nv: 🟠 **Major**: The PR is titled 'container CI handoff' but nothing runs this: `static-tests.yml` only runs pre-commit, `blossom-ci.yml` delegates to Jenkins with no reference to this script, and there's no runner label/job for a GPU host with pre-pulled nvstaging images. As merged it's a manual developer harness with hard host prerequisites. Either (a) wire it into a real GPU job (Jenkins/blossom) so 'CI' is true, or (b) retitle/document it as a local integration harness. Also rename: `scripts/test_*.py` is collected by `pytest scripts` (it has no tests, but the name implies it does) — something like `scripts/deft_container_handoff.py` matches the repo's `check_*`/`verify-*` naming.

scripts/test_dinov3_deft_containers.py
```diff
@@ -0,0 +1,251 @@
...
+    """Return command output without shell interpretation."""
+    return subprocess.check_output(argv, cwd=cwd, text=True).strip()
+
+
+def source_info(path):
+    """Require a clean committed snapshot and enumerate its feature diff."""
+    if capture(["git", "status", "--porcelain"], path):
+        raise ValueError(f"Commit or stash changes before testing: {path}")
+    commit = capture(["git", "rev-parse", "HEAD"], path)
+    parent = capture(["git", "rev-parse", "HEAD^"], path)
```
vpraveen-nv: 🟠 **Major**: `HEAD^` is hard-coded as the base for every repo. That only works for one-squashed-commit branches; on a normal multi-commit feature branch `changed_files`, the in-container diff check and `pre-commit --from-ref HEAD^` all silently cover just the last commit while still reporting success. Add a per-repo `--base-ref` (default `git merge-base origin/main HEAD`) and archive that instead.

scripts/test_dinov3_deft_containers.py
```diff
@@ -0,0 +1,251 @@
...
+def packaged_check(repo, require_gpu_faiss=False):
+    """Check the installed image, without importing a mounted source checkout."""
+    if repo == "tao-core":
+        module = importlib.import_module("nvidia_tao_core.config.dinov3.default_config")
+        config = module.ExperimentConfig()
+        if not hasattr(config.dataset, "train_manifest") or not hasattr(config, "grit_score"):
+            raise RuntimeError("Installed tao-core lacks the DINOv3 refinement schema")
+        print("PACKAGED_RUNTIME_CHECK_PASSED", flush=True)
+        return 0
+    modules = (["nvidia_tao_ds.mining.dinov3.entrypoint.refinement",
```
vpraveen-nv: 🟠 **Major**: Two problems with the packaged module lists. (1) For `repo == "tao-skill-bank"` this falls through to the tao-pytorch list and runs in the PyTorch image, so the skill-bank 'packaged' run is byte-for-byte the tao-pytorch run — nothing skill-specific is validated, and it's the wrong image (the skill's minimal profile is one *DS* container). (2) This list already diverges from DS `workflow/cli.py preflight` (`...ssl.dinov3.scripts.train` there vs `workflow.cli` / `entrypoint.dinov3` here). Suggest: skill-bank packaged check = run the DS image and execute `python -m nvidia_tao_ds.mining.dinov3.workflow preflight --gpu`, so there's exactly one definition of readiness.

scripts/test_dinov3_deft_containers.py
```diff
@@ -0,0 +1,251 @@
...
+               if repo == "tao-data-services" else [
+                   "nvidia_tao_pytorch.ssl.dinov3.data_refinement.cli",
+                   "nvidia_tao_pytorch.ssl.dinov3.data_refinement.train_cli",
+                   "nvidia_tao_pytorch.ssl.dinov3.entrypoint.dinov3",
+               ])
+    for name in modules:
+        imported = importlib.import_module(name)
+        print(json.dumps({"module": name, "path": imported.__file__}), flush=True)
+    import torch
+    if not torch.cuda.is_available():
```
vpraveen-nv: 🟡 **Minor**: `torch.cuda.is_available()` is the only GPU check. #245's `tao-container-integration.md` explicitly says allocation != readiness and makes the bf16/fp16 conv fwd/bwd probe the bar (noting rc-36 fails it). So this 'installed-image readiness' suite would pass on an image the skill docs call unusable. Call the DS CLI preflight (previous comment) or replicate the probe here.

scripts/test_dinov3_deft_containers.py
```diff
@@ -0,0 +1,251 @@
...
+        subprocess.run(["git", "-c", "user.name=Container CI", "-c", "user.email=ci@example.invalid",
+                        "commit", "--no-gpg-sign", "-qm", "CI candidate snapshot"], cwd=work, check=True)
+        observed = capture(["git", "diff", "--name-only", "--diff-filter=ACMR", "HEAD^", "HEAD"], work).splitlines()
+        if observed != manifest["repos"][repo]["changed_files"]:
+            raise RuntimeError("Archived snapshots do not reproduce the approved feature diff")
+        venv = root / "test-tools"
+        os.environ["PIP_CACHE_DIR"] = str(root / "pip-cache")
+        subprocess.run([sys.executable, "-m", "venv", "--system-site-packages", str(venv)], check=True)
+        python = str(venv / "bin/python")
+        subprocess.run([python, "-m", "pip", "install", "pytest", "pre-commit", "pylint", "pydocstyle", "flake8"], check=True)
```
vpraveen-nv: 🟠 **Major**: Unpinned `pip install pytest pre-commit pylint pydocstyle flake8` from PyPI inside the container at test time (plus pre-commit cloning hook repos). Requires network egress from the container, makes lint results non-reproducible across runs (same failure mode as the unpinned static-tests hook), and installs into the image's site-packages view via `--system-site-packages`. Pin versions (small constraints file under `inputs/`) and document that the source suite needs network; the packaged suite should stay offline.

scripts/test_dinov3_deft_containers.py
```diff
@@ -0,0 +1,251 @@
...
+        env = dict(os.environ)
+        env.update(PYTHONDONTWRITEBYTECODE="1", PRE_COMMIT_HOME=str(root / "pre-commit"),
+                   PIP_CACHE_DIR=str(root / "pip-cache"), TMPDIR=str(root),
+                   PATH=str(venv / "bin") + os.pathsep + env["PATH"],
+                   TAO_DATA_SERVICES_SOURCE=str(root / "tao-data-services"),
+                   PYTHONPATH=str(work), VALIDATE_BASE_REF="HEAD^",
```
vpraveen-nv: ❔ **Question**: `PYTHONPATH=str(work)` overlays the source tree on top of the installed package. For tao-pytorch that shadows compiled CUDA extensions living in site-packages (MSDeformAttn etc. aren't in the source tree), so any collected test that imports them fails for reasons unrelated to the change. The selected DINOv3 paths probably avoid this, but `tests/core` and `tests/distributed` are broad — have you confirmed they pass under overlay in the rc-36 image?

scripts/test_dinov3_deft_containers.py
```diff
@@ -0,0 +1,251 @@
...
+        return int(any(results))
+
+
+def run(args):
+    """Snapshot clean commits, pin local image IDs, and run all three checks."""
+    paths = dict(zip(REPOS, (args.skill_bank, args.pytorch, args.data_services)))
+    if args.core is not None:
+        paths["tao-core"] = args.core
+    metadata = {name: source_info(path.resolve()) for name, path in paths.items()}
+    images = {name: args.data_services_image if name == "tao-data-services" else args.pytorch_image for name in paths}
```
vpraveen-nv: 🟡 **Minor**: `tao-skill-bank` and `tao-core` are both mapped to `--pytorch-image`. tao-core in the PyT image is right (schema lives there); skill-bank should run in the DS image, which is the runtime the skill declares in `skill_info.yaml`.

scripts/test_dinov3_deft_containers.py
```diff
@@ -0,0 +1,251 @@
...
+        for suffix, ref in (("", "commit"), ("-base", "base")):
+            with (inputs / f"{name}{suffix}.tar").open("wb") as stream:
+                subprocess.run(["git", "archive", metadata[name][ref]], cwd=path, stdout=stream, check=True)
+    statuses = {}
+    for name in paths:
+        results = output / name
+        results.mkdir()
+        container = "deft-ci-" + uuid.uuid4().hex
+        argv = ["docker", "run", "--rm", "--pull=never", "--name", container,
+                "--gpus", args.gpus, "--shm-size=8g", "--entrypoint", "python",
```
vpraveen-nv: 🟡 **Minor**: Container runs as root (no `--user`), so container-written files (`regression.xml`/`checks.json`) come back root-owned under `--output`; `output.mkdir(exist_ok=False)` then makes a rerun with the same path impossible without sudo. Consider `--user $(id -u):$(id -g)` (release images have a passwd entry, unlike the dev base image) or a chown step. Nit: `--mount type=bind,src={inputs},...` breaks on paths containing `,`.

scripts/test_dinov3_deft_containers.py
```diff
@@ -0,0 +1,251 @@
...
+    parser = argparse.ArgumentParser(description=__doc__)
+    commands = parser.add_subparsers(dest="mode", required=True)
+    host = commands.add_parser("run")
+    for name in ("skill-bank", "pytorch", "data-services", "output"):
+        host.add_argument("--" + name, type=Path, required=True)
+    host.add_argument("--core", type=Path, help="Optional tao-core checkout for schema parity and lint")
+    host.add_argument("--pytorch-image", required=True)
+    host.add_argument("--data-services-image", required=True)
+    host.add_argument("--suite", choices=("source", "packaged"), default="source")
+    host.add_argument("--require-gpu-faiss", action="store_true",
```
vpraveen-nv: ⚪ **Nit**: `--require-gpu-faiss` is only meaningful with `--suite packaged`; with the default `--suite source` it's silently ignored (still appended to the child argv and recorded in `summary.json` as if checked). Error out on the combination or make the help text say packaged-only.

scripts/tests/test_dinov3_container_handoff.py
```diff
@@ -0,0 +1,105 @@
...
+    runs = [argv for argv in calls if argv[:2] == ["docker", "run"]]
+    removals = [argv for argv in calls if argv[:2] == ["docker", "rm"]]
+    assert len(runs) == len(removals) == (4 if include_core else 3)
+    assert not any("build" in argv or "pull" in argv for argv in calls)
+    for argv, cleanup in zip(runs, removals):
+        assert "--pull=never" in argv
+        assert argv[argv.index("--gpus") + 1] == "device=1"
+        assert any(part.startswith("sha256:") for part in argv)
+        assert cleanup[-1] == argv[argv.index("--name") + 1]
+        assert "CUDA_VISIBLE_DEVICES=0" not in argv
```
vpraveen-nv: 🟡 **Minor**: `assert "CUDA_VISIBLE_DEVICES=0" not in argv` asserts the absence of something the code never sets — it can't fail. Likewise `inside()`, ~60% of the script (archive extraction, synthetic git baseline, diff reproduction check, venv + pre-commit orchestration), has zero coverage while the trivially-mockable `run()` gets a 2x2x2 parametrisation. A tmp_path test that builds two tiny tarballs and drives `inside()` with `subprocess.run` patched only for pip/pre-commit/pytest would catch real regressions (e.g. the `.git` retention loop).

scripts/tests/test_dinov3_container_handoff.py
```diff
@@ -0,0 +1,105 @@
...
+
+import pytest
+
+MODULE = Path(__file__).resolve().parents[1] / "test_dinov3_deft_containers.py"
+SPEC = importlib.util.spec_from_file_location("deft_container_handoff", MODULE)
+handoff = importlib.util.module_from_spec(SPEC)
+SPEC.loader.exec_module(handoff)
+
+
+def args(tmp_path, dry_run=False):
```
vpraveen-nv: ⚪ **Nit**: `args()` omits `require_gpu_faiss`, so `run()` needs the `getattr(args, "require_gpu_faiss", False)` fallback purely to satisfy this fixture. Add the field here and drop the getattr — real argparse always sets it.

scripts/test_dinov3_deft_containers.py
```diff
@@ -0,0 +1,352 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Portable DEFT container checks using existing images, never Docker builds."""
```
vpraveen-nv, line 5: 🟠 **Major**: The PR is titled 'container CI handoff' but nothing runs this: `.github/workflows/static-tests.yml` only runs pre-commit, `blossom-ci.yml` delegates to Jenkins with no reference to this script, and there's no runner label/job for a GPU host with pre-pulled nvstaging images and clean sibling checkouts. As merged it's a manual developer harness. Either (a) wire it into a real GPU job so 'CI' is true, or (b) retitle and document it as a local integration harness (a short usage block in `docs/` or the module docstring). Also consider renaming: `scripts/test_*.py` implies a pytest module; something like `scripts/deft_container_handoff.py` matches the repo's `check_*`/`verify-*` naming.

scripts/test_dinov3_deft_containers.py
```diff
@@ -0,0 +1,352 @@
...
+        subprocess.run(["git", "-c", "user.name=Container CI", "-c", "user.email=ci@example.invalid",
+                        "commit", "--no-gpg-sign", "-qm", "CI candidate snapshot"], cwd=work, check=True)
+        observed = capture(["git", "diff", "--name-only", "--diff-filter=ACMR", "HEAD^", "HEAD"], work).splitlines()
+        if observed != manifest["repos"][repo]["changed_files"]:
+            raise RuntimeError("Archived snapshots do not reproduce the approved feature diff")
+        venv = root / "test-tools"
+        os.environ["PIP_CACHE_DIR"] = str(root / "pip-cache")
+        subprocess.run([sys.executable, "-m", "venv", "--system-site-packages", str(venv)], check=True)
+        python = str(venv / "bin/python")
+        subprocess.run([python, "-m", "pip", "install", "pytest", "pre-commit", "pylint", "pydocstyle", "flake8"], check=True)
```
vpraveen-nv, line 253: 🟠 **Major**: Unpinned `pip install pytest pre-commit pylint pydocstyle flake8` from PyPI inside the container at test time, plus pre-commit cloning hook repos on first run. This needs network egress from the container, makes lint results non-reproducible run to run (the same failure mode as the unpinned `static-tests` hook), and the `packaged` suite is otherwise offline. Please pin versions (a small constraints file copied into `inputs/` next to `driver.py`) and document that the `source` suite requires network while `packaged` does not.

scripts/test_dinov3_deft_containers.py
```diff
@@ -0,0 +1,352 @@
...
+
+
+def packaged_check(repo, require_gpu_faiss=False):
+    """Check the installed image, without importing a mounted source checkout."""
+    if repo == "tao-core":
+        core = importlib.import_module("nvidia_tao_core.config.dinov3.default_config")
+        grit = importlib.import_module("nvidia_tao_core.config.dinov3.grit_score")
+        runtime = importlib.import_module("nvidia_tao_pytorch.config.dinov3.default_config")
+        config = core.ExperimentConfig()
+        expected_grit = {
```
vpraveen-nv, line 120: 🟡 **Minor**: This is a hand-maintained copy of every `GRITScoreConfig` default plus `dataset`/`train` fields from tao-core #38. Any default change in tao-core (or a new field) now requires editing this skill-bank script, and the failure message will say 'installed tao-core lacks the complete schema' when the real cause is a legitimate default bump. The useful invariants are (1) the modules import, (2) `fields(core.GRITScoreConfig) == fields(runtime.GRITScoreConfig)` and (3) `asdict(core) == asdict(runtime)` (the parity tao-core's own test can't enforce because it skips there). Please drop the literal table and assert parity plus presence only.

vpraveen-nv (commented):
Reviewed at edef2e5. Adds `scripts/test_dinov3_deft_containers.py` — a host-side driver that `git archive`s clean checkouts of skill-bank/pytorch/data-services[/core] at their `merge-base origin/main`, pins pre-pulled image IDs, and runs either a `source` suite (PYTHONPATH overlay + pre-commit + selected pytest inside the image) or a `packaged` suite (module imports + DS `workflow preflight --gpu` + opt-in GPU FAISS probe) — plus host-only unit tests and a cross-repo dummy-data smoke that only runs when `nvidia_tao_ds` is importable. Good choices: real merge-base, readiness delegated to the DS CLI preflight (one definition), skill-bank/DS/core all in `--data-services-image`, no `shell=True`, `--pull=never`, image-ID pinning, `tarfile filter="data"`, forced cleanup on timeout. SPDX/DCO fine.

Remaining concerns: despite the `[CI]` title nothing invokes this (no workflow, Jenkinsfile, or hook), so it's a manual GPU developer harness with hard host prerequisites; the source suite does an unpinned `pip install` from PyPI plus pre-commit hook clones inside the container; the tao-core packaged check hard-codes a 21-field copy of the GRIT defaults that will drift with tao-core; `/results` comes back root-owned; `--require-gpu-faiss` is silently ignored on the default `source` suite; and `inside()` (the bulk of the logic) has no test. Comment / request changes: reposition as a dev harness (or wire it into a GPU job), pin the tool versions, and add an `inside()` test.

scripts/test_dinov3_deft_containers.py
```diff
@@ -0,0 +1,352 @@
...
+                + json.dumps(mismatches, sort_keys=True)
+            )
+        print("PACKAGED_RUNTIME_CHECK_PASSED", flush=True)
+        return 0
+    modules = (["nvidia_tao_ds.mining.dinov3.entrypoint.refinement",
+                "nvidia_tao_ds.mining.dinov3.workflow.cli",
+                "nvidia_tao_ds.mining.dinov3.workflow.native_actions",
+                "nvidia_tao_pytorch.ssl.dinov3.scripts.train",
+                "nvidia_tao_pytorch.ssl.dinov3.scripts.grit_score"]
+               if repo in {"tao-data-services", "tao-skill-bank"} else [
```
vpraveen-nv, line 178: 🟡 **Minor**: For `repo == "tao-skill-bank"` the packaged check is byte-for-byte the tao-data-services one (same module list, same `preflight --gpu`) in the same image, so the skill-bank container run adds wall time without validating anything skill-specific. Either skip skill-bank in `--suite packaged` or make it skill-specific, e.g. resolve `scripts/resolve_tao_image.py --application tao-run-dinov3-ssl-deft --action run` on the host and assert its image ID equals `--data-services-image` — that would have caught the #245 default-image mismatch.

scripts/test_dinov3_deft_containers.py
```diff
@@ -0,0 +1,352 @@
...
+    env = dict(os.environ)
+    env.pop("TAO_DEFT_REQUIRE_CROSS_REPO_SMOKE", None)
+    env.update(
+        PYTHONDONTWRITEBYTECODE="1",
+        PRE_COMMIT_HOME=str(root / "pre-commit"),
+        PIP_CACHE_DIR=str(root / "pip-cache"),
+        TMPDIR=str(root),
+        PATH=str(venv / "bin") + os.pathsep + env["PATH"],
+        TAO_DATA_SERVICES_SOURCE=str(root / "tao-data-services"),
+        PYTHONPATH=source_pythonpath(repo, root),
```
vpraveen-nv, line 100: ❔ **Question**: `PYTHONPATH=<source tree>` overlays the checkout on top of the installed package. For tao-pytorch that shadows any compiled extension living in site-packages but not in the source tree, so a collected test that imports one fails for reasons unrelated to the change. The description reports 323 passed in the PyT image, so the selected paths evidently avoid this today, but `tests/core` and `tests/distributed` are broad — is that overlay-safety documented anywhere, or should the selection be pinned to DINOv3-only paths to keep it that way?

scripts/test_dinov3_deft_containers.py
```diff
@@ -0,0 +1,352 @@
...
+    statuses = {}
+    for name in paths:
+        results = output / name
+        results.mkdir()
+        container = "deft-ci-" + uuid.uuid4().hex
+        argv = ["docker", "run", "--rm", "--pull=never", "--name", container,
+                "--gpus", args.gpus, "--shm-size=8g", "--entrypoint", "python",
+                "-e", "PYTHONDONTWRITEBYTECODE=1",
+                "--mount", f"type=bind,src={inputs},dst=/inputs,readonly",
+                "--mount", f"type=bind,src={results},dst=/results",
```
vpraveen-nv, line 307: 🟡 **Minor**: The container runs as root (no `--user`), so `regression.xml`/`checks.json` under `--output` come back root-owned and the host user can't clean the output tree without sudo. Consider `--user {os.getuid()}:{os.getgid()}` (release images have a passwd entry, unlike the dev base image) or a chown step at the end of `inside()`. Nit: `--mount type=bind,src={inputs},...` breaks on paths containing `,`.

scripts/test_dinov3_deft_containers.py
```diff
@@ -0,0 +1,352 @@
...
+    parser = argparse.ArgumentParser(description=__doc__)
+    commands = parser.add_subparsers(dest="mode", required=True)
+    host = commands.add_parser("run")
+    for name in ("skill-bank", "pytorch", "data-services", "output"):
+        host.add_argument("--" + name, type=Path, required=True)
+    host.add_argument("--core", type=Path, help="Optional tao-core checkout for schema parity and lint")
+    host.add_argument("--pytorch-image", required=True)
+    host.add_argument("--data-services-image", required=True)
+    host.add_argument("--suite", choices=("source", "packaged"), default="source")
+    host.add_argument("--require-gpu-faiss", action="store_true",
```
vpraveen-nv, line 338: ⚪ **Nit**: `--require-gpu-faiss` is only meaningful with `--suite packaged`; with the default `--suite source` it's silently ignored in `inside()` while still being appended to the child argv and recorded in `summary.json` as if checked. Error out on the combination or say packaged-only in the help text.

scripts/test_dinov3_deft_containers.py
```diff
@@ -0,0 +1,352 @@
...
+    """Return command output without shell interpretation."""
+    return subprocess.check_output(argv, cwd=cwd, text=True).strip()
+
+
+def source_info(path):
+    """Require a clean committed snapshot and enumerate its feature diff."""
+    if capture(["git", "status", "--porcelain"], path):
+        raise ValueError(f"Commit or stash changes before testing: {path}")
+    commit = capture(["git", "rev-parse", "HEAD"], path)
+    base = capture(["git", "merge-base", "origin/main", commit], path)
```
vpraveen-nv, line 35: ⚪ **Nit**: Good change to `merge-base origin/main`. This now hard-requires a remote literally named `origin` tracking `main` in every checkout (the `--core` checkout included); a `--base-ref` override (default `origin/main`) would make it usable from devbox checkouts whose remotes are still named `gitlab`, and from release branches.

scripts/tests/test_dinov3_container_handoff.py
```diff
@@ -0,0 +1,210 @@
...
+    runs = [argv for argv in calls if argv[:2] == ["docker", "run"]]
+    removals = [argv for argv in calls if argv[:2] == ["docker", "rm"]]
+    assert len(runs) == len(removals) == (4 if include_core else 3)
+    assert not any("build" in argv or "pull" in argv for argv in calls)
+    for argv, cleanup in zip(runs, removals):
+        assert "--pull=never" in argv
+        assert argv[argv.index("--gpus") + 1] == "device=1"
+        assert any(part.startswith("sha256:") for part in argv)
+        assert cleanup[-1] == argv[argv.index("--name") + 1]
+        assert "CUDA_VISIBLE_DEVICES=0" not in argv
```
vpraveen-nv, line 202: 🟡 **Minor**: `assert "CUDA_VISIBLE_DEVICES=0" not in argv` asserts the absence of something the code never sets; it can't fail. More importantly `inside()` (archive extraction, synthetic git baseline, diff reproduction check, venv + pre-commit orchestration — roughly 60% of the script) has zero coverage while the trivially-mockable `run()` gets a 2×2×2 parametrisation. A `tmp_path` test that builds two tiny tarballs plus a `manifest.json`, points `/inputs` and `/results` at temp dirs (make them module constants), and drives `inside()` with `subprocess.run` patched only for pip/pre-commit/pytest would catch real regressions in the `.git` retention loop and the diff-reproduction check.

scripts/tests/test_dinov3_container_handoff.py
```diff
@@ -0,0 +1,210 @@
...
+
+import pytest
+
+MODULE = Path(__file__).resolve().parents[1] / "test_dinov3_deft_containers.py"
+SPEC = importlib.util.spec_from_file_location("deft_container_handoff", MODULE)
+handoff = importlib.util.module_from_spec(SPEC)
+SPEC.loader.exec_module(handoff)
+
+
+def args(tmp_path, dry_run=False):
```
vpraveen-nv, line 22: ⚪ **Nit**: `args()` omits `require_gpu_faiss`, so `run()` needs the `getattr(args, "require_gpu_faiss", False)` fallback at line 277 purely to satisfy this fixture. Add the field here and drop the getattr; real argparse always sets it.

scripts/tests/test_dinov3_cross_repo_smoke.py
```diff
@@ -0,0 +1,206 @@
...
+        build_runner,
+        client_job_id,
+    )
+    from nvidia_tao_pytorch.ssl.dinov3.dataloader.dataset import DinoV3Dataset
+    from nvidia_tao_pytorch.ssl.dinov3.utils.refinement_attestation import (
+        canonical_digest,
+        file_sha256,
+        native_attestation,
+    )
+except ModuleNotFoundError as error:
```
vpraveen-nv, line 36: 🟡 **Minor**: Only `ModuleNotFoundError` is turned into a module-level skip. If a host has an older `nvidia_tao_ds`/`nvidia_tao_pytorch` installed (e.g. the released 7.2.0 packages) that lacks `_native_implementation_files` or `refinement_attestation`, the `from ... import` raises a plain `ImportError`, and `pytest scripts/tests` errors at collection instead of skipping. Catch `ImportError` (the parent class) here, and keep the `TAO_DEFT_REQUIRE_CROSS_REPO_SMOKE=1` re-raise so the DS handoff still fails hard.

scripts/tests/test_dinov3_cross_repo_smoke.py
```diff
@@ -0,0 +1,206 @@
...
+    skill_schema = json.loads((
+        skill_root / "skills/applications/tao-run-dinov3-ssl-deft/"
+        "references/workflow.schema.json"
+    ).read_text(encoding="utf-8"))
+    import nvidia_tao_ds.mining.dinov3.workflow as workflow_package
+    runtime_schema = json.loads((
+        Path(workflow_package.__file__).resolve().parent /
+        "schemas/workflow.schema.yaml"
+    ).read_text(encoding="utf-8"))
+    assert runtime_schema == skill_schema
```
vpraveen-nv, line 192: 🟡 **Minor**: Good to see a drift test for the schema copy. Two gaps: it covers only `workflow.schema`, while `metrics.schema.json` and `stage-request.schema.json` in #245 are equally verbatim copies; and it reads the DS `.yaml` with `json.loads`, which works today only because the DS files happen to be written in JSON syntax. Please loop over all three and parse the DS side with `yaml.safe_load`.

scripts/test_dinov3_deft_containers.py
nnagrajrao: Corrected the PR title to explicitly say manual cross-repository DEFT container test harness. The tool is scripts/deft_container_handoff.py, and its module docstring documents usage and the source/package validation boundaries. This is a manual integration harness, not a wired Jenkins job.

Code revision remains 2a62c866; title correction requires no code change.

scripts/test_dinov3_deft_containers.py
nnagrajrao: Pinned the test-tool versions in TEST_TOOLS and documented source-mode network requirements versus offline packaged checks. Repository-pinned pre-commit hooks run unchanged with no SKIP exemptions; this is not a hermetic dependency lock for production packages.

Revision: `2a62c866`.

scripts/test_dinov3_deft_containers.py
nnagrajrao: Removed the literal default table. Packaged validation now checks required feature presence and dataclass field/default/metadata parity between installed Core and runtime schemas.

Revision: `2a62c866`.

scripts/deft_container_handoff.py
```diff
@@ -0,0 +1,352 @@
...
+                + json.dumps(mismatches, sort_keys=True)
+            )
+        print("PACKAGED_RUNTIME_CHECK_PASSED", flush=True)
+        return 0
+    modules = (["nvidia_tao_ds.mining.dinov3.entrypoint.refinement",
+                "nvidia_tao_ds.mining.dinov3.workflow.cli",
+                "nvidia_tao_ds.mining.dinov3.workflow.native_actions",
+                "nvidia_tao_pytorch.ssl.dinov3.scripts.train",
+                "nvidia_tao_pytorch.ssl.dinov3.scripts.grit_score"]
+               if repo in {"tao-data-services", "tao-skill-bank"} else [
```
nnagrajrao: Packaged mode now skips Skill Bank; it is not an installed TAO runtime package. The separate skill release-image mismatch is explicitly blocked in #245.

Revision: `2a62c866`.

scripts/deft_container_handoff.py
```diff
@@ -0,0 +1,352 @@
...
+    env = dict(os.environ)
+    env.pop("TAO_DEFT_REQUIRE_CROSS_REPO_SMOKE", None)
+    env.update(
+        PYTHONDONTWRITEBYTECODE="1",
+        PRE_COMMIT_HOME=str(root / "pre-commit"),
+        PIP_CACHE_DIR=str(root / "pip-cache"),
+        TMPDIR=str(root),
+        PATH=str(venv / "bin") + os.pathsep + env["PATH"],
+        TAO_DATA_SERVICES_SOURCE=str(root / "tao-data-services"),
+        PYTHONPATH=source_pythonpath(repo, root),
```
nnagrajrao: Documented the source-overlay boundary explicitly: these selected Python tests do not validate wheel contents or compiled-extension compatibility. Kept shared core/distributed coverage to detect collateral changes; packaged probes remain separate.

Revision: `2a62c866`.

scripts/deft_container_handoff.py
```diff
@@ -0,0 +1,352 @@
...
+    statuses = {}
+    for name in paths:
+        results = output / name
+        results.mkdir()
+        container = "deft-ci-" + uuid.uuid4().hex
+        argv = ["docker", "run", "--rm", "--pull=never", "--name", container,
+                "--gpus", args.gpus, "--shm-size=8g", "--entrypoint", "python",
+                "-e", "PYTHONDONTWRITEBYTECODE=1",
+                "--mount", f"type=bind,src={inputs},dst=/inputs,readonly",
+                "--mount", f"type=bind,src={results},dst=/results",
```
nnagrajrao: Containers now run with the host UID/GID, and comma-containing bind paths fail early with a clear error. The new output artifacts are user-owned.

Revision: `2a62c866`.

scripts/deft_container_handoff.py
```diff
@@ -0,0 +1,352 @@
...
+    parser = argparse.ArgumentParser(description=__doc__)
+    commands = parser.add_subparsers(dest="mode", required=True)
+    host = commands.add_parser("run")
+    for name in ("skill-bank", "pytorch", "data-services", "output"):
+        host.add_argument("--" + name, type=Path, required=True)
+    host.add_argument("--core", type=Path, help="Optional tao-core checkout for schema parity and lint")
+    host.add_argument("--pytorch-image", required=True)
+    host.add_argument("--data-services-image", required=True)
+    host.add_argument("--suite", choices=("source", "packaged"), default="source")
+    host.add_argument("--require-gpu-faiss", action="store_true",
```
nnagrajrao: The source/require-gpu-faiss combination now errors instead of recording a check that never ran; help describes the packaged-only flag.

Revision: `2a62c866`.

scripts/test_dinov3_deft_containers.py
nnagrajrao: Added --base-ref with origin/main as default; validation used an explicitly fetched GitHub main reference.

Revision: `2a62c866`.

scripts/tests/test_dinov3_container_handoff.py
nnagrajrao: Added an inside() test using real tiny archives and a synthetic git baseline, mocking only tool invocations. It caught a same-stat index-reuse bug; emptying the temporary index before staging now makes diff reproduction reliable.

Revision: `2a62c866`.

scripts/tests/test_dinov3_container_handoff.py
nnagrajrao: The fixture supplies require_gpu_faiss and the runtime no longer needs the getattr fallback.

Revision: `2a62c866`.

scripts/tests/test_dinov3_cross_repo_smoke.py
nnagrajrao: Catch ImportError for optional standalone collection; TAO_DEFT_REQUIRE_CROSS_REPO_SMOKE=1 still reraises so the combined DS harness fails closed.

Revision: `2a62c866`.

scripts/tests/test_dinov3_cross_repo_smoke.py
nnagrajrao: The smoke now checks all three schemas and parses the DS side with yaml.safe_load.

Revision: `2a62c866`.

scripts/deft_container_handoff.py
```diff
@@ -0,0 +1,352 @@
...
+        for suffix, ref in (("", "commit"), ("-base", "base")):
+            with (inputs / f"{name}{suffix}.tar").open("wb") as stream:
+                subprocess.run(["git", "archive", metadata[name][ref]], cwd=path, stdout=stream, check=True)
+    statuses = {}
+    for name in paths:
+        results = output / name
+        results.mkdir()
+        container = "deft-ci-" + uuid.uuid4().hex
+        argv = ["docker", "run", "--rm", "--pull=never", "--name", container,
+                "--user", f"{os.getuid()}:{os.getgid()}",
```
vpraveen-nv, line 302: 🟡 **Minor**: `--user uid:gid` is passed without `-e HOME=/tmp -e USER=... -e LOGNAME=...`. In the release images the host UID has no passwd entry, so `HOME` resolves to `/`: (a) `import torch` in PyTorch 2.11+ calls `getpass.getuser()` during dynamo cache init and can crash the packaged probes; (b) with `SKIP` now empty the repos' `language: golang` trufflehog hook needs a writable GOCACHE/HOME and will fail in the source suite. #245's own `tao-container-integration.md` sets exactly these env vars for the same reason. The description's validation predates the `--user`/`SKIP` changes, so this combination hasn't been exercised in a container yet — please add the env and re-run one source + one packaged container.

vpraveen-nv (approved):
Verified all 12 round-2 items at 2a62c86: renamed to `scripts/deft_container_handoff.py` and documented as a manual harness; pinned tool versions; GRIT literal table replaced with field/`asdict` parity; skill-bank dropped from the packaged matrix; `--user`; `--require-gpu-faiss` guard; `--base-ref`; a real `inside()` test (I reproduced the new path locally with real git); fixture field; `ImportError` catch; three-schema `yaml.safe_load` drift test. DCO/static green. Approving.

This can't merge until #245 lands (base is the draft controller branch) — retarget to `main` if #245 stays draft for long. One new item worth fixing before the next container run, inline. Also drop the now-dead `tao-skill-bank` packaged branches (lines ~165/174) and refresh 'How was this patch tested' with results from the current head.

PR #250, opened 2026-09-15
fix(rtdetr): support explicit cold-start baseline

echen4628:

Add an explicit `baseline_mode` contract to the DEFT RT-DETR workflow with two supported values:

- `cold_start`, the default.
- `checkpoint`.

When `prepare_deft_od_aoi_measurement.py` is invoked with `--baseline` in `cold_start` mode, it:

- Projects KPI ground truth into the expected KITTI representation.
- Creates an empty prediction file for every KPI image.
- Emits only the loose and strict gap-analysis specs.
- Does not emit or run baseline KPI/test inference.

This makes every initial KPI ground-truth box a false negative for the first retrieval iteration.

In explicit `checkpoint` mode, baseline behavior remains checkpoint-based: the helper emits KPI and test inference specs plus both gap-analysis specs.

Measurements after training are unchanged. They omit `--baseline` and always run inference using the selected binary checkpoint.

Example:

```bash
scripts/prepare_deft_od_aoi_measurement.py \
  --policy "$RESULTS/deft_od_aoi_policy.yaml" \
  --checkpoint "$CHECKPOINT" \
  --kpi-predictions "$MEASURE/kpi/inference/labels" \
  --results-root "$MEASURE" \
  --output-dir "$MEASURE/specs" \
  --baseline
```

The workflow previously ran baseline inference using the training initializer. In deployments where that initializer has a multiclass head but DEFT uses a binary background/defect contract, treating it as a compatible binary baseline is incorrect and can fail or produce misleading initial gaps.

A true cold start should represent the state before a compatible detector has been trained. Empty baseline predictions model that state directly while still allowing the frozen checkpoint to initialize iteration training.

The explicit mode also prevents the workflow from silently guessing whether a supplied checkpoint is binary-compatible.

### Related issues

N/A

Yes.

Previously, the baseline always emitted KPI and test inference specs using the supplied base checkpoint.

The default `cold_start` mode now skips baseline inference and initializes gap analysis with empty KPI predictions. Users who have a binary-compatible checkpoint and want the previous checkpoint-based baseline can set:

```yaml
baseline_mode: checkpoint
```

Existing hand-authored policies must add either `baseline_mode: cold_start` or `baseline_mode: checkpoint`. Policies copied from the updated default template receive `cold_start` automatically.

Training initialization is unchanged: every iteration still starts from the configured trainable base checkpoint.

Focused initialization and measurement tests were run from an isolated checkout of this branch:

```console
$ .venv/deft/bin/python -m pytest -q \
    skills/applications/tao-run-deft-od-aoi/scripts/tests/test_init_deft_od_aoi.py \
    skills/applications/tao-run-deft-od-aoi/scripts/tests/test_prepare_deft_od_aoi_measurement.py

..........                                                               [100%]
10 passed in 0.56s
```

The tests cover:

- The default cold-start mode.
- Explicit checkpoint mode.
- Rejection of unsupported automatic mode selection.
- Cold-start empty KPI prediction generation.
- Omission of baseline inference specs during cold start.
- Checkpoint-based baseline inference.
- Normal inference after the baseline, regardless of baseline mode.

The cold-start path was also exercised in a three-iteration acceptance run that completed training, KPI/test measurement, gap analysis, and final completion validation.

Yes. This patch was co-authored with OpenAI Codex.

Generated-by: OpenAI Codex (GPT-5)

### Release note

```release-note
Added an explicit cold-start RT-DETR baseline that initializes DEFT gaps without requiring inference from an incompatible multiclass checkpoint.
```

### Checklist

- [x] All tests pass locally
- [x] I added or updated documentation (README, docstrings, docs pages, examples)
- [x] I updated the version, changelog, or migration notes if this change requires it
- [x] If this touches components that are optional to install, the imports are guarded so the package still works without them (reviewers: please verify this)
- [x] I understand this contribution is licensed under the repository's license, and that my commit author name and email become permanently public once merged

### Notes for reviewers

The key behavioral branch is in `prepare_deft_od_aoi_measurement.py`.

Please verify the intentional separation between:

- Baseline measurement mode, which determines whether initial inference runs.
- Training initialization, which continues to use the configured base checkpoint.
- Later iteration measurements, which always use the selected binary checkpoint.

The generated `measurement_manifest.json` is authoritative about which inference specs exist; callers should submit only the specs listed there.


arihantj-nv:
why do we need a cold start mode? would we not always train from some checkpoint given to us from TAO?

echen4628:
Training-wise, we do start from the TAO checkpoint. For RT-DETR, the checkpoint used here is `rtdetr_warehouse_v1.0.2.pth`, which was trained with `num_classes: 7`. During training, TAO reuses the compatible pretrained weights and initializes new task-specific classification layers for our binary background/defect setup.

However, the first gap analysis occurs before any training. Using the supplied checkpoint at that point would mean running inference directly with its seven-class head, which does not match the binary task.

Cold-start mode avoids treating that incompatible checkpoint as a valid binary baseline. If a user instead has a binary-compatible checkpoint and wants to use it for the initial inference, they can set `baseline_mode: checkpoint`.

xiaosean (approved):

arihantj-nv (approved):
synced offline

PR #251, opened 2026-09-15
fix(deft): parse split RT-DETR KPI events

echen4628:

Update DEFT’s RT-DETR status parser to support KPI and epoch information arriving in separate JSON events.

The parser now:

- Tracks the latest valid epoch event.
- Associates a later untagged finite `val_mAP50` or `mAP50` KPI with that epoch.
- Continues to prefer an epoch carried directly by the KPI object or event.
- Avoids double-counting an untagged KPI when RT-DETR immediately repeats the same value in the next epoch-tagged event.
- Ignores malformed, nonnumeric, and non-finite KPI values as before.

For example, these two events are now interpreted as epoch 7 with `val_mAP50` of `0.61`:

```json
{"epoch": 7, "step": 120}
{"kpi": {"val_mAP50": 0.61}}
```

RT-DETR can emit the current epoch and evaluation KPI in separate status events. The previous parser required both values to appear in the same event, so it could discard valid evaluation results and report that the status contained no finite KPI rows.

That prevented DEFT from reliably comparing training probes, updating KPI history, and selecting a checkpoint even though RT-DETR had emitted the required metric.

RT-DETR may also repeat an untagged evaluation value in the following epoch-tagged event. Handling that pattern prevents the same evaluation from being recorded twice or attributed to the wrong epoch.

### Related issues

N/A

Yes.

Previously, DEFT could fail checkpoint or probe selection when RT-DETR emitted epoch and KPI data in separate events.

It now reconstructs the metric’s epoch from the event stream and proceeds with selection. No command, configuration, or data migration is required.

Focused training-selection tests were run from an isolated checkout of this branch:

```console
$ .venv/deft/bin/python -m pytest -q \
    skills/applications/tao-run-deft-od-aoi/scripts/tests/test_select_deft_od_aoi_training.py

....                                                                     [100%]
4 passed in 0.56s
```

New tests cover:

- Carrying an epoch forward to a later KPI-only event.
- Preferring the following epoch-tagged event when it repeats the same KPI value.
- Avoiding duplicate KPI history entries.

The existing probe-selection and checkpoint-history tests continue to pass.

The split-event parser was also exercised during a three-iteration RT-DETR acceptance run that completed probe selection, main and extension training, measurement, and final completion validation.

Yes. This patch was co-authored with OpenAI Codex.

Generated-by: OpenAI Codex (GPT-5)

### Release note

```release-note
Fixed DEFT checkpoint selection when RT-DETR reports epoch and validation KPI data in separate status events.
```

### Checklist

- [x] All tests pass locally
- [x] Documentation is unchanged because this internal parser fix does not change a command or configuration contract
- [x] No version, changelog, or migration update is required
- [x] This change does not add or alter optional imports
- [x] I understand this contribution is licensed under the repository's license, and that my commit author name and email become permanently public once merged

### Notes for reviewers

The review can focus on `_metrics()` in `select_deft_od_aoi_training.py` and its two new event-sequence tests.

The pending untagged KPI logic is deliberately narrow: it only carries a KPI when a valid preceding epoch exists, and it suppresses that pending value only when the next tagged event repeats the same score.


skills/applications/tao-run-deft-od-aoi/scripts/select_deft_od_aoi_training.py
```diff
@@ -24,16 +24,51 @@ def _metrics(paths: list[Path]) -> list[tuple[int, float]]:
...
         if not path.is_file():
             raise FileNotFoundError(path)
+        latest_epoch = None
+        pending_untagged = None
         for line in path.read_text(errors="ignore").splitlines():
             try:
                 row = json.loads(line)
             except json.JSONDecodeError:
                 continue
+            row_epoch = None
```
arihantj-nv, line 34: What is the purpose of the commit? since DEFT mostly focuses on end of training results what does help with?

echen4628:
I think there are two actually separate questions here:

1. Does DEFT need to know the best epoch, or is the final epoch sufficient?
2. Why is this commit specifically needed for RT-DETR?

For the first question, the current workflow needs the best epoch for several reasons:

1. The checkpoint from the epoch with the highest validation AP50 is used for the next gap analysis—not necessarily the checkpoint from the final epoch.
2. When the optional probing/HPO phase is enabled, each candidate runs for 10 epochs. We use its best AP50 across the full probe to compare candidates and select the configuration for the main training run.
3. We report the best AP50 achieved during model training.

Therefore, only using the final training result would change the current selection policy and could select a worse checkpoint if performance peaked earlier.

This commit is needed because RT-DETR can emit the epoch and its validation AP50 in separate status events. The existing parser expected them in the same event, so it could fail to associate an AP50 value with its epoch even after training completed. This change reconstructs that association from the completed event history so DEFT can correctly select and report the best epoch.

Does this answer the question?

arihantj-nv (approved):
okay this is more of a reporter fix.

PR #260, opened 2026-09-16
[TAO-2230][Bugfix] deft-od: carry the declared AP50 gates into gap_analysis

arihantj-nv:
Fixes NVBug 6776554 (P0, `deft-release-blocker`).

## Why

`init_deft_state.py` accepts per-class AP50 thresholds, validates them, and stores them at `config.ap50_thresholds`. `gap_analysis` needs them as a `weak_thresholds` block, and `apply_spec_overrides.py` can only set a nested block from a file.

Nothing produced that file. The documented build named it as:

```bash
--set-from-file weak_thresholds="<the threshold mapping from the section above>"
```

`--set-from-file` takes a path; that is prose. So the one line carrying the operator's gates could not be executed. The stage fell back to the gates hardcoded in `assets/gap_analysis_object_detection.yaml`, exited 0, and warned about nothing. The flag validated carefully and had no consumer.

It stays invisible whenever a run happens to declare the same values the asset holds, which is why it survived several full runs.

**The gates are not only a reporting choice.** The weak set sizes the mining budget (`weak_count × multiplier`), so substituted gates change which images the iteration mines.

## What

`prepare_thresholds_for_gap_analysis.py` writes `config.ap50_thresholds` into the file the spec build consumes, in `{class: {ap50: value}}` shape.

It reads the state the run already froze rather than taking the thresholds again on the command line — a second source is a second thing to disagree with `deft_state.json`. It re-checks what init checked, because this file is what the stage actually reads: a target class with no gate would be scored against `default_ap50_threshold: 0.0`, never marked weak, and never mined for.

The reference now shows the producer before the build, and says what happens without it.

## Verification

The reporter's own case, which previously ended `equal: NO`:

```
declared : {car 0.50, bicycle 0.90, person 0.90}
state    : {'car': 0.5, 'bicycle': 0.9, 'person': 0.9}
spec     : {'bicycle': {'ap50': 0.9}, 'car': {'ap50': 0.5}, 'person': {'ap50': 0.9}}
equal    : YES
```

842 state-machine assertions, up from 836. New G28 covers the round-trip with values deliberately far from the asset's, and the refusal when a target class has no gate.

## Note

The suggested fix in the bug warns against deleting the line and relying on "the emitted spec already holds every documented default". That is not what this does — the default path is unchanged; the operator's value now reaches the stage.


echen4628 (approved):

skills/applications/tao-run-deft-object-detection/scripts/prepare_thresholds_for_gap_analysis.py
```diff
@@ -0,0 +1,113 @@
...
+
+        for name, value in thresholds.items():
+            if isinstance(value, bool) or not isinstance(value, (int, float)):
+                raise ValueError(f"ap50_thresholds[{name!r}] is {value!r}, not a number")
+            if not 0 <= float(value) <= 1:
+                raise ValueError(f"ap50_thresholds[{name!r}] is {value}, outside [0, 1]")
+
+        # Target-class order, so the file reads the way the run declares its classes.
+        ordered = [c for c in targets if c in thresholds]
+        ordered += [c for c in thresholds if c not in ordered]
```
vpraveen-nv, line 91: This also writes gates for non-target classes. init warns those "will be ignored", but once they're in this file they aren't: gap_analysis marks an image weak if *any* listed class is under its gate, so a stray `truck: 0.7` inflates the weak set and the mining budget. Write `targets` only (or fail on extras):

```python
block = {"weak_thresholds": {c: {"ap50": float(thresholds[c])} for c in targets}}
```

skills/applications/tao-run-deft-object-detection/scripts/tests/test_state_machine.sh
```diff
@@ -2495,6 +2495,51 @@ assert_eq '0.3' "$(state_json "$G27/results/run_g27" kpi_conf_threshold)" \
...
+person: [person]"
+G28_RUN="$G28/results/run_g28"
+# Deliberately far from the asset's 0.99/0.7/0.7 so a match cannot be coincidental.
+init_run "$G28" "$G28_RUN" 1 --target-classes bicycle,car,person \
+  --ap50-thresholds-json '{"car": 0.50, "bicycle": 0.90, "person": 0.90}'
+assert_rc 0 "[G28] init accepts and stores the declared gates"
+
+run "$PY" "$SCRIPTS_DIR/prepare_thresholds_for_gap_analysis.py" \
+  --results-dir "$G28_RUN" --out "$G28_RUN/weak_thresholds.yaml"
+assert_rc 0 "[G28] the gates are written to a file the spec build can consume"
```
vpraveen-nv, line 2520: Worth one more assertion that actually runs the build line this fixes: `apply_spec_overrides.py --spec <asset copy> --set-from-file weak_thresholds=$G28_RUN/weak_thresholds.yaml`, then check the output has only the run's classes (use a target set that isn't car/bicycle/person). That pins the single-key unwrap and replace-not-merge behaviour the fix depends on. I ran it by hand and it works; it just isn't covered.

vpraveen-nv (changes requested):
Diagnosis and fix shape are right: the file comes from frozen state, and `set_path` replaces the whole `weak_thresholds` leaf (I checked — the asset's car/bicycle/person get swapped out, not merged). One real fix needed (non-target classes leak into the file) and one test ask, inline.

Doc nit outside the diff: `pipeline-and-state.md` step 1 still says "see the overlay" — there is no gap overlay; name `prepare_thresholds_for_gap_analysis.py` there like the KPI step names its prepare script.

Heads-up: #262 and #274 also add a `G28` block at the same line of `test_state_machine.sh`; whichever lands second needs renumbering.

skills/applications/tao-run-deft-object-detection/scripts/prepare_thresholds_for_gap_analysis.py
arihantj-nv: Fixed in b4c15477. The file now carries `config.target_classes` only, in the run's declared order, and names anything dropped on stderr. You're right that it was worse than a stray entry: it was the same class of defect the PR fixes, gates the operator didn't aim at the run sizing its mining budget.

skills/applications/tao-run-deft-object-detection/scripts/tests/test_state_machine.sh
arihantj-nv: Added, and it caught more than I expected. G30 builds the spec through the real `--set-from-file weak_thresholds=…` line with a `bus` target and an untargeted `truck` gate in state, then asserts the built spec holds only `bus` at the declared 0.6. So it pins the single-key unwrap and replace-not-merge. With the scoping fix reverted, `truck` reaches the built spec, not just the intermediate file.

arihantj-nv:
@vpraveen-nv all three addressed. The non-target gates are dropped, and G30 now runs the `--set-from-file` build line. `pipeline-and-state.md` step 1 names `prepare_thresholds_for_gap_analysis.py` instead of the overlay that doesn't exist. Rebased onto #274 with the gate renumbered to G30. #262 took G29 and #261 took G31, so none of the three needs renumbering again, whatever order they land in.

skills/applications/tao-run-deft-object-detection/references/tao-analyze-gaps-od-map.md
```diff
@@ -51,9 +51,25 @@ Build it the same way every other stage does — emit, then fill the run-specifi
   --set images_dir="$KPI_IMAGES_DIR" \
   --set results_dir="${RESULTS_DIR}/iter${N}/gaps" \
   --set kpi="iter${N}" \
-  --set-from-file weak_thresholds="<the threshold mapping from the section above>"
+  --set-from-file weak_thresholds="${RESULTS_DIR}/iter${N}/gaps/weak_thresholds.yaml"
```
nnagrajrao: The file this reads is written by the block just below. Could you swap the order?

nnagrajrao (approved):
Good catch. Reading from frozen state is the right fix, and scoping to target classes makes sense.

Follow-ups:
- `weak_thresholds: ???` in the asset would make skipping the step fail loudly.
- gap_analysis matches class names case-sensitively with no mapping, so KPI GT with aliases could mark nothing weak. Worth an issue.
- nit: "the emitted spec already holds every documented default" and the YAML example in `tao-analyze-gaps-od-map.md` still suggest car/bicycle/person are the spec.

Approving to unblock.

skills/applications/tao-run-deft-object-detection/references/tao-analyze-gaps-od-map.md
arihantj-nv: Swapped in 942460a7. The prepare step now comes first, as "First, write the run's gates to a file", with the build after it. Your nit is in the same commit: the doc no longer says the emitted spec "already holds every documented default" (it says the asset's car/bicycle/person gates are a placeholder that file replaces), and the YAML example shows one entry per target class.

arihantj-nv:
@nnagrajrao thanks. On the two follow-ups:

- **`weak_thresholds: ???`**: done in 75dd0467. The documented build also passes `--require-no-mandatory` now, so skipping the prepare step fails at the build and names the field. With the old asset the new assertions fail, because the skip is accepted silently. That's the path this bug took.
- **Case-sensitive matching**: confirmed in the pinned DS image. `gap_analysis` applies `class_mapping.get(cls, cls)` to both GT and predictions, and the loop always passes `{}`, while `kpi_analyze` folds aliases through its mapping. So on a KPI set whose ground truth uses aliases (`Car`, `Automobile`, `rider`…), KPI and gap analysis see different classes, and the aliased class is never gated. The ITS KPI set used so far is labelled with the canonical names, so existing runs aren't affected. It's a separate defect from this one, so I'll track it separately rather than grow this PR.

PR #261, opened 2026-09-16
[TAO-2230][Bugfix] deft-od: fetch and check the Co-DETR checkpoint

arihantj-nv:
Fixes NVBug 6759013 (P1).

## Why

Prep pseudo-labels the entire pool with Co-DETR. The skill declares `CODETR_CHECKPOINT` as a required run variable and passes it to `init_deft_state.py` — then never obtains it and never checks for it. None of the Pre-Flight checks looked at it, no bundled script fetched it, and the only pointer was a line in `prep-source-pool.md` sitting *below* the command that consumes the variable, naming a different skill.

The zero-shot Grounding DINO checkpoint is handled the opposite way: a numbered Pre-Flight check, a fetch script with `--plan`, idempotent reuse. Two required checkpoints, one fetched and verified by the skill, the other left to the reader.

That asymmetry isn't arbitrary — `assets/overlays/codetr_inference.yaml` pins the ViT-L/16 geometry that exactly one published checkpoint loads. The skill had already chosen the answer and simply didn't fetch it.

The cost lands late: the Co-DETR pass is the longest stage in the workflow (~2.5 h on a 100k pool) and it is the first thing prep does.

## What

**`fetch_codetr_checkpoint.py`** — downloads `zongzhuofan/co-detr-vit-large-coco` over **stdlib HTTPS**.

The documented route was `huggingface-cli`, which QA showed is not installable on a clean box (`externally-managed-environment`) and whose entry point is now `hf`. Requiring `huggingface_hub` instead would have put a network-fetching package into `deft_python.sh`'s interpreter probe — making one stage's dependency a precondition for every run, including those that never label a pool, and making the reported problem *worse*. `urllib` needs nothing.

Integrity is checked, not assumed: size and SHA-256 verified before the file is moved into place, downloading to a `.part` first. A checkpoint that doesn't match the pinned architecture loads no weights, exits 0, prints `Execution status: PASS`, and writes one empty label file per image — so a wrong or truncated file is refused rather than warned about.

**Pre-Flight check 7**, which runs only when prep will run; later checks renumbered and every `check N` cross-reference updated to match.

**Prep's Paths block** exports the checkpoint beside the other paths, instead of having the variable first appear inside the launch. **Prep's Gates** list a missing checkpoint first.

## Verification

| case | result |
|---|---|
| `--plan`, empty dest | reports, downloads nothing (0 files) |
| `--plan`, already present | `ALREADY PRESENT` + size |
| checkpoint present | reused, prints path, exit 0 |
| correct SHA-256 | downloads, keeps, prints path |
| wrong SHA-256 | exit 1, **0 files left** — no `.part`, no truncated file |
| one char changed in the digest | exit 1, 0 files left |

Source URL and size confirmed live: `2,934,763,233` bytes, matching the copy this workflow has labelled pools with since August. The pinned SHA-256 was computed from that copy; HuggingFace's ETag is the xet object id, not a content hash, so it can't stand in for it.

836 state-machine assertions unchanged; all bash blocks parse.


echen4628 (approved):

skills/applications/tao-run-deft-object-detection/scripts/fetch_codetr_checkpoint.py
```diff
@@ -0,0 +1,169 @@
...
+                        break
+                    handle.write(chunk)
+                    digest.update(chunk)
+                    written += len(chunk)
+                    if declared:
+                        print(f"\r  {_human(written)} / {_human(declared)}",
+                              end="", file=sys.stderr, flush=True)
+        if declared:
+            print(file=sys.stderr)
+    except urllib.error.URLError as exc:
```
vpraveen-nv, line 111: Only `URLError` cleans up the `.part`. A timeout on `response.read()`, `IncompleteRead`, `ConnectionResetError` or Ctrl-C all leave a partial file behind, which contradicts "0 files left" in the description and makes the free-space check fail on retry. Use `try/finally` to unlink `part` unless it was moved into place.

skills/applications/tao-run-deft-object-detection/scripts/fetch_codetr_checkpoint.py
```diff
@@ -0,0 +1,169 @@
...
+                print(f"ALREADY PRESENT: {target} ({_human(target.stat().st_size)})",
+                      file=sys.stderr)
+            else:
+                print(f"WILL DOWNLOAD after approval: {REPO}/{FILENAME} "
+                      f"({_human(EXPECT_BYTES)}) -> {target}", file=sys.stderr)
+            return 0
+
+        # Idempotent, and checked rather than trusted: a half-written file from an
+        # interrupted transfer is the same size problem as a wrong checkpoint.
+        if target.is_file():
```
vpraveen-nv, line 142: Reuse checks size only. Reasonable trade for 2.8 GB, but the docstring says "checked rather than trusted" and same-size-wrong-bytes is what the SHA is for. A `--verify` flag (off by default) would let the agent opt in on a suspicious run. Not blocking.

skills/applications/tao-run-deft-object-detection/references/prep-source-pool.md
```diff
@@ -18,6 +18,8 @@ mkdir -p "$PREP_DIR"
 
 export CLASSES_YAML="<path to the run's classes.yaml>"
 export POOL_IMAGES="<config.pool_images>"   # the raw images this prep labels
+export CODETR_CHECKPOINT="<from Pre-Flight check 7>"  # the pseudo-labeller's weights
```
vpraveen-nv, line 21: init already records this as `config.codetr_checkpoint`. Use `<config.codetr_checkpoint>` here, consistent with the classmap line below and with #260 reading from state.

skills/applications/tao-run-deft-object-detection/references/preflight.md
```diff
@@ -121,16 +121,44 @@ Resolve everything you can before asking the user. Parameter precedence is stric
...
+   on a file somebody staged by hand.
+
+   ```bash
+   # Report first; --plan prints a sentence, not a path.
+   <skill_root>/scripts/deft_python.sh \
+     <skill_root>/scripts/fetch_codetr_checkpoint.py --plan \
+     --dest "$WORKSPACE/checkpoints/codetr"
+
+   # After approval, the same command without --plan prints the path.
+   CODETR_CHECKPOINT=$(<skill_root>/scripts/deft_python.sh \
```
vpraveen-nv, line 137: Worth one sentence on why `--dest` is under `$WORKSPACE`: the Co-DETR container only mounts `$WORKSPACE` + `$EXTRA_MOUNTS`, so anywhere else needs an extra `-v`.

vpraveen-nv (commented):
Good change — stdlib-only is right given the `deft_python.sh` probe, and `.part` + verify + `replace()` is the correct pattern. Comments inline. Outside the diff: `init_deft_state.py` L18 still says "check 9" for the encoder, now check 10.

Before merge I'd like a test. `urllib` accepts `file://`, so `--url file:///$G/small.bin --expect-sha256 <sha>` (and a wrong digest) covers mismatch / no-files-left / reuse in the existing harness with no network.

skills/applications/tao-run-deft-object-detection/scripts/fetch_codetr_checkpoint.py
arihantj-nv: Fixed. Cleanup is now in `finally`, guarded on whether the file was moved into place, so a timeout, `IncompleteRead`, a reset or Ctrl-C all leave nothing. The `file://` tests you suggested are in G31, but worth knowing: all of them pass against the old code too, because it already unlinked on `URLError` and on a digest mismatch. A failure *after* bytes arrive is the leak, and `file://` can't produce one. So the assertion that pins it calls `download()` with a response that dies on its second read. That one fails with the fix reverted.

skills/applications/tao-run-deft-object-detection/scripts/fetch_codetr_checkpoint.py
arihantj-nv: Left for a follow-up, as you suggested. Same-size-wrong-bytes on reuse is real but narrower than the leak, and I'd rather not grow this PR.

skills/applications/tao-run-deft-object-detection/references/prep-source-pool.md
arihantj-nv: Done, now `<config.codetr_checkpoint>`.

skills/applications/tao-run-deft-object-detection/references/preflight.md
arihantj-nv: Added, just after the intro to check 7.

arihantj-nv:
@vpraveen-nv cleanup moved to `finally` with a test that fails without it, and the three doc points are done: check 10, `config.codetr_checkpoint`, and why `--dest` sits under `$WORKSPACE`. `--verify` is in too (2db073dd), with a same-size-wrong-bytes test. Rebased onto #274 with the gate at G31.

skills/applications/tao-run-deft-object-detection/scripts/fetch_codetr_checkpoint.py
arihantj-nv: Changed my mind on deferring this. It's in now (2db073dd). `--verify` hashes a reused checkpoint against `--expect-sha256`, off by default since it reads 2.8 GiB. On a mismatch it refuses and leaves the file for the operator rather than deleting it, and it refuses `--verify` with an empty digest rather than treating that as a pass. G31 covers the case the flag exists for: one byte flipped at the same length, reused silently without `--verify` and refused with it.

skills/applications/tao-run-deft-object-detection/references/preflight.md
```diff
@@ -121,16 +121,53 @@ Resolve everything you can before asking the user. Parameter precedence is stric
...
+   on every launch that reads it.
+
+   ```bash
+   # Report first; --plan prints a sentence, not a path.
+   <skill_root>/scripts/deft_python.sh \
+     <skill_root>/scripts/fetch_codetr_checkpoint.py --plan \
+     --dest "$WORKSPACE/checkpoints/codetr"
+
+   # After approval, the same command without --plan prints the path.
+   CODETR_CHECKPOINT=$(<skill_root>/scripts/deft_python.sh \
```
nnagrajrao: This replaces a checkpoint the user may already have staged, which blocks air-gapped hosts. Could it follow the gdino pattern (use the user's path if given, else fetch)?

skills/applications/tao-run-deft-object-detection/scripts/fetch_codetr_checkpoint.py
```diff
@@ -0,0 +1,214 @@
...
+
+def main() -> int:
+    try:
+        args = parse_args()
+        dest = Path(args.dest).expanduser().resolve()
+        target = dest / FILENAME
+
+        if args.plan:
+            if target.is_file():
+                print(f"ALREADY PRESENT: {target} ({_human(target.stat().st_size)})",
```
nnagrajrao: Could `--plan` do the size check too? Otherwise a truncated file shows as present in the summary.

skills/applications/tao-run-deft-object-detection/references/preflight.md
```diff
@@ -121,16 +121,53 @@ Resolve everything you can before asking the user. Parameter precedence is stric
    image the checkpoint was trained with in the Summary; the pinned image is not
    automatically the right one.
 
-7. **Train-spec template.** Must exist and parse as YAML, and `dataset.train_data_sources` must be a **list** (Grounding DINO ODVG shape). A mapping there means the spec is COCO-shaped and this workflow cannot append to it.
+7. **Co-DETR checkpoint — only when `prep` will run.** Prep pseudo-labels the whole
```
nnagrajrao: Worth a grep for "check N" references outside the skill too.

nnagrajrao (approved):
Stdlib-only makes sense here, and the `.part` + verify + `finally` handling looks right.

The main open item is the user-supplied checkpoint case. A follow-up is fine.

Approving to unblock.

skills/applications/tao-run-deft-object-detection/references/preflight.md
arihantj-nv: Done in d3de3aa6. Check 7 now follows check 6: the user's path wins and the fetcher is skipped. It must exist. A size different from the published checkpoint is flagged in the Summary rather than refused, since a compatible fine-tune can differ. A path outside `$WORKSPACE` has to be in `$EXTRA_MOUNTS`.

skills/applications/tao-run-deft-object-detection/scripts/fetch_codetr_checkpoint.py
arihantj-nv: Done in d3de3aa6. The reuse checks (size, plus the digest with `--verify`) are now one function that `--plan` and reuse both call, so they can't drift apart. A truncated file at `--dest` now fails the plan (exit 1) instead of showing as ALREADY PRESENT. G31 covers it with the default URL, which `--plan` never contacts, so it runs offline. The three new assertions fail without the change.

skills/applications/tao-run-deft-object-detection/references/preflight.md
arihantj-nv: Checked. No reference outside the skill cites a Pre-Flight check number. Inside it, every one points at the right check (3 credentials, 5 image presence, 7 Co-DETR, 9 source pool, 10 encoder). One wording fix in d3de3aa6: prep's gate said check 7 "fetches" the checkpoint, and it now says it resolves it, from the user's path or a fetch.

PR #262, opened 2026-09-16
[TAO-2230][Bugfix] deft-od: say which kind of complete a finished run was

arihantj-nv:
Fixes NVBug 6776880 (P1).

## Why

A run that finished every iteration and a run that stopped early on a documented condition both record `status=complete`. `audit_deft_run.py` computes `completion_reason` to separate them, and deliberately prints it on success as well as failure — its own comment says the reason is the only thing that tells the two apart.

But printing is not an artifact. `deft_state.json` carried no reason, so on a run that hard-stopped at iteration 3 of a permitted 4 the report read:

```
**Status:** COMPLETE
**Iterations completed:** 2 / 4
```

with nothing saying it was truncated or why. `2 / 4` doesn't carry it either — the counts differ by design, since an iteration that stopped part-way is started but not completed.

There was also a straight contradiction inside the skill. `agents/reporter.md` said *"the audit is the authority: a run stopped early records `stopped`, never `complete`"*, while `audit_deft_run.py` has returned `complete=true` for a documented early stop since that behaviour was added. Whichever was intended, the other was wrong.

## What

**The semantics stay.** An exhausted pool has done everything there was to do, so it is a completed run, not an abandoned one. The documentation is what changes.

- `commit_stage.py` writes `completion_reason` into state, beside the `status` it already takes from the same audit report.
- `agents/reporter.md` now says a run can be `COMPLETE` without running every iteration, and requires the reason beside the status whenever it names an early stop:

```
**Status:** COMPLETE (documented early stop: the source pool was exhausted at iter3)
**Iterations completed:** 2 / 4
```

The reporter's existing `STOPPED (INCOMPLETE)` branch is untouched — it still covers a genuinely abandoned run.

## Verification

New G28 drives a full run to a pool-exhaustion stop after one completed iteration and asserts both halves: `status` is still `complete`, and `completion_reason` names the exhaustion. 839 assertions, up from 836.

## Note

This does not change `--require-complete`. A pool-exhaustion stop passed before and passes now, so runs QA has already verified stay verified.


echen4628 (approved):

skills/applications/tao-run-deft-object-detection/agents/reporter.md
```diff
@@ -48,11 +48,30 @@ Take **Status** from the audit's `--json` report, never from prose and never fro
...
+`deft_state.json`'s `status` field mirrors that verdict (`running`, `stopped`, `complete`,
+`failed`) and the audit is the authority for it.
+
+**A run can be `COMPLETE` without having run every iteration.** Two documented early stops
+— an exhausted source pool, and an iteration that found no weak images — end the loop
+having done everything there was to do, so the audit calls them complete rather than
+abandoned. They are not the same run as one that finished all `max_iterations`, and
+`completion_reason` is the only thing that tells them apart.
+
+So whenever `completion_reason` names an early stop, print it beside the status rather
```
vpraveen-nv, line 64: This asks the reporter to pattern-match prose ("documented early stop:"). The audit already knows this structurally — a `completion_kind` enum next to the string (`all_iterations | zero_weak_images | pool_exhausted | incomplete`) would let it branch without string matching. Follow-up is fine.

skills/applications/tao-run-deft-object-detection/scripts/commit_stage.py
```diff
@@ -980,6 +980,14 @@ def main() -> int:
             # wrong on its own.
             if stage == TERMINAL_STAGE and args.status == "ok" and not run_failed:
                 state["status"] = "complete" if report.get("complete") else "stopped"
+                # A run that finished every iteration and a run that stopped early on a
+                # documented condition both record status=complete, and the reason is the
+                # only thing that separates them. The audit computes it and prints it, but
+                # printing is not an artifact: without this the state file says a truncated
+                # run completed and offers nothing to say why.
+                reason = report.get("completion_reason")
```
vpraveen-nv, line 988: `completion_reason` is always non-empty so `if reason:` never fails, but if it did a stale reason would sit next to a fresh `status`. `state["completion_reason"] = reason or None` keeps them together. Note this also writes the reason for `stopped` runs, which is useful — worth a word in reporter.md's STOPPED row.

vpraveen-nv (approved):
LGTM. Keeping the semantics and fixing the docs contradiction is right, and taking the reason from the same audit report as `status` keeps them in lockstep. Two non-blocking notes inline. Same `G28` collision with #260/#274 in `test_state_machine.sh`.

skills/applications/tao-run-deft-object-detection/scripts/commit_stage.py
arihantj-nv: Done in 0c36355e: `state["completion_reason"] = report.get("completion_reason") or None`, written beside `status` from the same report. For the record, per the audit's branches every terminal `loop_stop` gets a non-empty reason today, so this guards against a future audit change rather than fixing a reachable case. The STOPPED row in `reporter.md` now prints the reason, and G29 pins that a hand-stopped run records `only 0 of 3 iterations finished…`.

skills/applications/tao-run-deft-object-detection/agents/reporter.md
arihantj-nv: Agreed. Filtering on the `"documented early stop:"` prefix is fragile. Leaving the `completion_kind` enum for a follow-up as you suggested.

PR #273, opened 2026-09-16
[TAO-2230][Bugfix] deft-od: stop asking a platform question nothing can act on

arihantj-nv:
Fixes NVBug 6763307 (P1).

## Why

Launch Intake asked the user which of the four installed platform skills they intended to run on — then stored the answer nowhere and branched on it never. Selecting SLURM, Kubernetes or Brev silently produced a Docker run, and the user wasn't told their choice had been discarded.

Measured on the skill as it stands:

```
platform refs in any script      0
platform field in state schema   0
submit / sbatch / kubectl / brev 0
unconditional `docker run`       22   (across 8 reference files)
```

There is no code path the answer could take.

## What

The bank already has the rule, at `AGENTS.md:51` — where an application supports one platform, treat that contract as the selection, and ask only when it supports several. This workflow supports one: its `compatibility` line requires docker plus the nvidia-container-toolkit, and it ships no `skill_info.yaml` declaring platforms.

So the question is **removed rather than implemented**, and the credentials step points at `tao-run-on-docker` directly.

Supporting the other three is a different change, and would need the selection recorded in state and dispatched through the bank's four-verb job contract.

One file, docs only. 836 assertions unchanged.


vpraveen-nv (approved):
LGTM. Removing rather than implementing is right — `compatibility` names one platform, no `skill_info.yaml` declares more, and AGENTS.md already says single platform = the selection.

PR #274, opened 2026-09-16
[TAO-2230][Bugfix] deft-od: refuse a dropped exclude set however it is dropped

arihantj-nv:
Fixes NVBug 6786515 (P1).

## Why

Iterations stay disjoint because each mine excludes everything mined before it. `prepare_exclude_for_mining.py` guarded that — but only halfway:

```python
if args.parquet_b:              # false when the flag is omitted → block skipped
    ...
    elif args.iteration == 1:
        print("NOTE: no previous cumulative at iteration 1")
    else:
        raise FileNotFoundError(...)   # only fires for a *bad path*
```

So `--parquet-b /wrong/path` was refused, and `--parquet-b` left off entirely was not. In the second case every earlier exclusion is dropped, the next mine is free to re-select images the model already trained on, and nothing in the output says so.

Worth noting the existing guard's comment already stated exactly this cost — it just wasn't reached on the omitted-flag path.

## What

Omitting the flag and pointing it at a missing file are the same mistake with the same cost, so they get the same guard. Only iteration 1 may do either.

The check moved to the top of `main()`, **before any parquet is opened**. It's argument consistency rather than a file check, so a run should be refused on its arguments rather than after reading this iteration's mined parquet — and that placement also makes it testable on a host with no parquet engine, which the test harness's interpreter is.

## Verification

| case | before | after |
|---|---|---|
| iter 2, flag omitted | **exit 0, exclusions dropped** | exit 1 |
| iter 2, flag → missing file | exit 1 | exit 1 |
| iter 1, flag omitted | exit 0 | exit 0 |
| iter 2, both given | exit 0, union carried | exit 0, union carried (3 rows from 2+1) |

The union and iteration-1 success paths were checked manually against real parquet files; the suite covers the refusal semantics, which is where the bug was. 841 assertions, up from 836.


echen4628 (approved):

skills/applications/tao-run-deft-object-detection/scripts/prepare_exclude_for_mining.py
```diff
@@ -44,6 +44,19 @@ def main() -> int:
     args = parser.parse_args()
 
     try:
+        # Argument consistency first, before any file is read: omitting the flag and
+        # pointing it at a missing file are the same mistake with the same cost, and
+        # only the second was caught. `if args.parquet_b:` is false when the flag is
+        # absent, which skipped the block and dropped every earlier exclusion without
+        # a word, leaving the next mine free to re-select images already trained on.
+        if not args.parquet_b and args.iteration != 1:
```
vpraveen-nv, line 52: The remaining way to drop the exclude set is `--iteration 1` passed on a later iteration (the error text even suggests it). State knows the truth — current phase and `iterations[iter{N-1}].exclude_parquet`. A follow-up taking `--results-dir` and deriving both from state, like #260's prepare script, would remove the retyped inputs. Not for this PR.

vpraveen-nv (approved):
LGTM. The omitted-flag hole was real and checking arguments before any file read is right. One follow-up thought inline. Same `G28` collision with #260/#262 in `test_state_machine.sh`.

PR #275, opened 2026-09-16
[TAO-2230][Bugfix] deft-od: keep the training val metrics out of the loop report

arihantj-nv:
## Why

Fixes NVBug 6777131 (P1).

`DEFT_Loop_Report.md` puts two numbers in the same stage table, two rows apart, that measure different things:

```
train          val_mAP50 = 0.8266
kpi_analyze    mAP       = 0.7841
```

`val_mAP50` scores the trained model against the validation split of the staged mined data. Those labels are Co-DETR pseudo-labels, so it measures how closely the student reproduces the teacher. `mAP` scores against the KPI evaluation set and its human ground truth — the accuracy figure the loop exists to move.

The names differ by one prefix, and on the measured run the agreement metric was the higher of the two, so the confusion runs in the direction that overstates the model. The report is what a customer reads to decide whether the loop worked.

The string `val_mAP` does not appear in the skill: the training container prints it, and a `train` summary that improvises from that stdout carries it into `loop_log.jsonl`, which the stage timeline reproduces verbatim.

## What

The report should not show the number at all, rather than show it with a legend — there is one accuracy metric in this loop and it comes from `kpi_analyze`.

`agents/reporter.md`:

- Section 3 drops any `val_` metric from a summary, reproducing the rest verbatim. The metric stays in `loop_log.jsonl` for anyone who wants it.
- Section 1 states that its numbers are scored against the KPI evaluation set, and that they are the only mAP in the report.
- "the iteration with the best mAP" in Observations now says best KPI mAP.
- The `kpi_conf_threshold` rule said "every mAP in the report is scored at it" — true once no other mAP is rendered, and now written that way.

`references/grounding-dino.md`: the train step's prescribed `--summary` was already `trained iter${N}: <epochs> epochs, <N> data sources` and never asked for a metric. Says so explicitly now, with the reason, so the number stops entering the log in the first place.

Docs only — no script, schema or state change. State machine tests pass 836/836.


echen4628 (approved):

skills/applications/tao-run-deft-object-detection/agents/reporter.md
```diff
@@ -105,20 +109,31 @@ lacked annotations for some images. Call that gap out when it is non-zero.
 One row per `loop_log.jsonl` event: seq, iter, stage, status, duration, summary.
 Include the `tokens` column only when the field is present.
 
+**Drop any `val_` metric a summary carries.** A `train` summary should read
```
vpraveen-nv, line 112: This relies on two agents following prose. Cheaper to enforce: have `commit_stage.py --stage train` reject or strip a `--summary` matching `val_mAP`, so the metric never enters `loop_log.jsonl` and the reporter has nothing to filter. Follow-up sized.

vpraveen-nv (approved):
LGTM. Dropping the number rather than adding a legend is right — there's one accuracy metric in this loop. One suggestion inline.

